PR #111, opened 2015-01-14
Update year in copyright license: Trello #693

spanktar:
The year in the license file needed to be updated from 2014 to 2015


mfussenegger:
what about the NOTICE file?

chaudum:
ack

PR #115, opened 2015-02-03
added missing functionality to cratedialect

spanktar:
Add missing dialect attributes from initialize method.
Add tests for the same.


mfussenegger:
changes are missing, otherwise ack

PR #117, opened 2015-02-13
change sqla update statement generation to not include table name

mfussenegger:


bogensberger:
ack, please also update admin-ui version in crate-0.47 after release

PR #119, opened 2015-02-17
fixes #118 - match literals

### LINKED ISSUE #118: sqlalchemy MATCH support is not properly escaped
Adding a single quote to a query utilizing MATCH will break the query. I modified the test case to induce the behavior:

```
    def test_simple_match(self):
        query = self.session.query(self.Character.name) \
                    .filter(match(self.Character.name, 'Trill\'ian'))
        self.assertSQL(
            "SELECT characters.name AS characters_name FROM characters WHERE match(characters.name, 'Trill\\'ian')",
            query
        )
```

Result:

```
Failure
Expected :"SELECT characters.name AS characters_name FROM characters WHERE match(characters.name, 'Trill\\'ian')"
Actual   :"SELECT characters.name AS characters_name FROM characters WHERE match(characters.name, 'Trill'ian')"
```

The output of the test error is a little misleading - the comparison is of course taking place without the escapes.

jeffnappi:


chaudum:
hi @jeffnappi 
LGTM
can you please add a changelog entry in the `CHANGES.txt` file, then I can merge it :)

jeffnappi:
@chaudum done :rocket:

Issue #116, opened 2015-02-04
Sqlalchemy support for NOT NULL


zopyx:
The Kotti table specs seems to define a NOT NULL constraint. This causes the following error:

Traceback (most recent call last):
  File "bin/pserve", line 11, in <module>
    sys.exit(main())
  File "/Users/ajung/src/kotti/lib/python2.7/site-packages/pyramid/scripts/pserve.py", line 51, in main
    return command.run()
  File "/Users/ajung/src/kotti/lib/python2.7/site-packages/pyramid/scripts/pserve.py", line 316, in run
    global_conf=vars)
  File "/Users/ajung/src/kotti/lib/python2.7/site-packages/pyramid/scripts/pserve.py", line 340, in loadapp
    return loadapp(app_spec, name=name, relative_to=relative_to, **kw)
  File "/Users/ajung/src/kotti/lib/python2.7/site-packages/paste/deploy/loadwsgi.py", line 247, in loadapp
    return loadobj(APP, uri, name=name, **kw)
  File "/Users/ajung/src/kotti/lib/python2.7/site-packages/paste/deploy/loadwsgi.py", line 272, in loadobj
    return context.create()
  File "/Users/ajung/src/kotti/lib/python2.7/site-packages/paste/deploy/loadwsgi.py", line 710, in create
    return self.object_type.invoke(self)
  File "/Users/ajung/src/kotti/lib/python2.7/site-packages/paste/deploy/loadwsgi.py", line 203, in invoke
    app = context.app_context.create()
  File "/Users/ajung/src/kotti/lib/python2.7/site-packages/paste/deploy/loadwsgi.py", line 710, in create
    return self.object_type.invoke(self)
  File "/Users/ajung/src/kotti/lib/python2.7/site-packages/paste/deploy/loadwsgi.py", line 146, in invoke
    return fix_call(context.object, context.global_conf, **context.local_conf)
  File "/Users/ajung/src/kotti/lib/python2.7/site-packages/paste/deploy/util.py", line 55, in fix_call
    val = callable(*args, **kw)
  File "/Users/ajung/src/kotti/lib/python2.7/site-packages/kotti/__init__.py", line 165, in main
    initialize_sql(engine)
  File "/Users/ajung/src/kotti/lib/python2.7/site-packages/kotti/resources.py", line 782, in initialize_sql
    metadata.create_all(engine, tables=tables)
[...]

PR #120, opened 2015-02-19
support table creation using the sqlalchemy ORM

mfussenegger:


jeffnappi:
What about ObjectArray field definitions and index options such as analyzer, etc?

mfussenegger:
Good point. I've actually played around with object field definitions at our last snow sprint (the rest of this PR is actually also a result of that).  But I've left that out as I wasn't really happy with how it worked - it is a bit trick because of the whole change tracking stuff.

I've made this PR anyway without that stuff so that at least the simple cases as mentioned in the issue here https://github.com/crate/crate-python/issues/116 will work.

src/crate/client/sqlalchemy/types.py
```diff
@@ -108,6 +108,9 @@ class Comparator(sqltypes.TypeEngine.Comparator):
         def __getitem__(self, key):
             return self._binary_operate(self.expr, operators.getitem, key)
 
+    def get_col_spec(self):
+        return 'object'
```
chaudum: a little picky: `OBJECT` because all other keywords are uppercase too.
(changes the output of `create_table_test.py:53`

chaudum:
otherwise ack

jeffnappi:
@mfussenegger sounds good, right now I have some of my own hand-rolled DDL generation. It would be nice to have something more robust.

PR #129, opened 2015-08-05
Fix urllib3 error with invalid kwargs for HTTPConnectionPool when REQ…

alexrad-spotify:
…UESTS_CA_BUNDLE is set

### DIFF
--- a/src/crate/client/http.py
+++ b/src/crate/client/http.py
@@ -159,7 +159,16 @@ class Client(object):
     def _update_server_pool(self, servers, **kwargs):
         for server in servers:
             if not server in self.server_pool:
-                self.server_pool[server] = Server(server, **kwargs)
+                https = server.lower().startswith('https:')
+                if not https:
+                    kwargs_copy = {}
+                    #clean up any kwargs which are invalid for HTTPConnectionPool
+                    for key in kwargs:
+                        if key not in ['ca_certs', 'cert_reqs']:
+                            kwargs_copy[key] = kwargs[key]
+                    self.server_pool[server] = Server(server, **kwargs_copy)
+                else:
+                    self.server_pool[server] = Server(server, **kwargs)
 
     @staticmethod
     def _server_url(server):
--- a/src/crate/client/test_http.py
+++ b/src/crate/client/test_http.py
@@ -22,6 +22,7 @@
 import json
 import time
 import sys
+import os
 from .compat import queue
 from random import SystemRandom
 import traceback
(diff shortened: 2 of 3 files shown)


src/crate/client/http.py
```diff
@@ -128,11 +128,13 @@ def __init__(self, servers=None, timeout=None, ca_cert=None,
         self._inactive_servers = []
         self._http_timeout = timeout
         pool_kw = {}
-        if ca_cert is None:
-            ca_cert = os.environ.get("REQUESTS_CA_BUNDLE", None)
-        if ca_cert is not None:
-            pool_kw['ca_certs'] = ca_cert
-            pool_kw['cert_reqs'] = verify_ssl_cert and 'REQUIRED' or 'NONE'
+        https = not any(['http:' in uri.lower() for uri in servers])
```
dobe: the check is actually wrong in cases where you have both http and https uris. if at least one https url exists it should match so you need to check the positive case. also you should check just the beginning of the url like this: 

```
https =  any([uri.lower().startswith('https:') for uri in servers])
```

dobe:
thanks a lot for you PR! could you pls sing the CLA, so we can merge it  (see https://crate.io/community/contribute/agreements/)

alexrad-spotify:
should be set

src/crate/client/http.py
alexrad-spotify: good points, hmm, in the mixed http+https case this will need refactoring to handle the two variants

alexrad-spotify:
Updated patch is up

src/crate/client/http.py
```diff
@@ -159,6 +159,12 @@ def _get_sql_path(self, **url_params):
     def _update_server_pool(self, servers, **kwargs):
         for server in servers:
             if not server in self.server_pool:
+                https = server.lower().startswith('https:')
+                if not https:
+                    #clean up any kwargs which are invalid for HTTPConnectionPool
+                    for key in ['ca_certs', 'cert_reqs']:
+                        if key in kwargs:
+                            del kwargs[key]
```
mfussenegger: I think this still won't work if there are https and non-https server in the list since the kwargs is shared.
You'd have to copy it once for http and once for https.

mfussenegger:
Would it also be possible to create a test for this?

chaudum:
LGTM

please could you add a changelog entry (in `CHANGES.txt`, underneath the `Unreleased` headline) and squash the commits into a single one.
then I can merge the PR

thanks!

alexrad-spotify:
Updated

src/crate/client/test_http.py
```diff
@@ -423,3 +424,15 @@ def test_params(self):
     def test_no_params(self):
         client = Client(['127.0.0.1:4200'])
         self.assertEqual(client.path, "_sql")
+
+
+class RequestsCaBundleTest(TestCase):
+
+    def test_open_client(self):
+        print 'o hai'
```
chaudum: oh, hi!
think this sneaked in again ;)

alexrad-spotify:
fixed the 'o hai'

chaudum:
thanks!

chaudum:
@alexrad-spotify made a new release `0.13.4` which is now available on pypi
https://pypi.python.org/pypi/crate/

PR #130, opened 2015-08-26
Client certificate support

mike-schiller:
Allows for optional client certificate / key to be used when connecting to a Crate server via https. This is useful if Crate is running behind a revers proxy that validates client certificates or for modified versions of Crate that support client certificate validation.

### DIFF
--- a/src/crate/client/connection.py
+++ b/src/crate/client/connection.py
@@ -110,6 +110,10 @@ def connect(servers=None, timeout=None, client=None,
     :param ca_cert:
         a path to a CA certificate to use when verifying the SSL server
         certificate.
+    :param cert_file:
+        a path to the client certificate to present to the server.
+    :param key_file:
+        a path to the client key to use when communicating with the server.
     :param error_trace:
         if set to ``True`` return a whole stacktrace of any server error if
         one occurs
--- a/src/crate/client/connection.py
+++ b/src/crate/client/connection.py
@@ -28,7 +28,9 @@ from distutils.version import StrictVersion
 
 class Connection(object):
     def __init__(self, servers=None, timeout=None, client=None,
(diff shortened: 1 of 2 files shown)


src/crate/client/http.py
```diff
@@ -127,7 +128,7 @@ def __init__(self, servers=None, timeout=None, ca_cert=None,
         self._active_servers = servers
         self._inactive_servers = []
         self._http_timeout = timeout
-        pool_kw = {}
+        pool_kw = {'cert_file':cert_file, 'key_file':key_file}
```
mfussenegger: does it really make sense to add them to the dict even if they're None?

src/crate/client/connection.py
```diff
@@ -28,7 +28,9 @@
 
 class Connection(object):
     def __init__(self, servers=None, timeout=None, client=None,
-                 verify_ssl_cert=False, ca_cert=None, error_trace=False):
+                 verify_ssl_cert=False, ca_cert=None,
+                 cert_file=None, key_file=None,
```
mfussenegger: could you add the args as last params in order to not break backward compatibility for people who might have used positional arguments?

mfussenegger:
Sorry for the late response. Seems like this PR was overlooked.

chaudum:
hi @mike-schiller 

do you want to add the changes @mfussenegger suggested?
also, could you add an entry to the CHANGES.txt file (in the "Unreleased" section)

thanks,
christian

ps: did you sign a [Contributor Agreement](https://crate.io/community/contribute/agreements/) already?

chaudum:
@mike-schiller updates on this one?

mfussenegger:
We've released 0.15 which contains client certificate support, so I'm closing this.

PR #132, opened 2015-10-04
add python3.5 to travis

mfussenegger:
somehow the tox recipe doesn't work on travis under
python3.5. I tried to fix that with
https://github.com/gawel/gp.recipe.tox/pull/1 and while that
does fix things for me locally it broke the travis build for
3.3, 3.4 and 3.5 (3.3 is odd because the behaviour for 3.3
is actually unchanged. 3.4 and 3.5 broke because pip is
missing. Which is odd too because `venv.create(..,
with_pip=True)` should ensure that it is there. Which is
actually the case locally, just not on travis).

To work around all that hassle I simply split up the
buildout configurations to not install the tox and sphinx
recipes in the travis run as they're not needed anyway.

### DIFF
--- a/.travis.yml
+++ b/.travis.yml
@@ -4,13 +4,14 @@ python:
   - 2.7
   - 3.3
   - 3.4
+  - 3.5
 sudo:
     false
 install:
   - pip uninstall -y setuptools
   - python bootstrap.py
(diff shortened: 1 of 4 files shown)


mfelsche:
i'll delegate the review process to mr. travis

PR #133, opened 2015-10-08
Make sqlalchemy Object column policies configurable

ckrybus:
Right now it is not easily possible to use a different column policy when using the Object type. This commit makes it easier and is backwards compatible.

Due to as_mutable() always returning an instance, it is not possible (to my knowledge) to extend the Object type "the sqlalchemy way" eg.

```python
# default Object column
object1 = Column(Object)

# throws TypeError because Object is already an instance
object2 = Column(Object('dynamic'))   
```

My solution is a little verbose, extends the api, the naming is also not optimal. if it is too much, you could keep only the changes from the _Craty class and they would still make configuring column policies way easier eg.

```python
# complete solution
object1 = Column(StrictObject)

# minimal change solution
object2 = Column(MutableDict.as_mutable(_Craty('strict')))   
```

### DIFF
--- a/src/crate/client/sqlalchemy/tests/create_table_test.py
+++ b/src/crate/client/sqlalchemy/tests/create_table_test.py
@@ -1,7 +1,7 @@
 import sqlalchemy as sa
 from sqlalchemy.ext.declarative import declarative_base
 
-from crate.client.sqlalchemy.types import Object
+from crate.client.sqlalchemy.types import IgnoredObject, Object, StrictObject
 from crate.client.cursor import Cursor
 
 from mock import patch, MagicMock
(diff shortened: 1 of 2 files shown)


mfussenegger:
Thanks for the contribution.

I'm not sure if it is a good idea to have explicit classes for each column_policy.
Instead I could imagine having a factory method. Something like:

    def object(column_policy):
        return MutableDict.as_mutable(_Craty(column_policy))

But I think before adding something like that we should take a look at how we can support specifying nested columns.
It should eventually be possible to specify something like e.g.

    create table t (
        obj (strict) as (
            name string
        )
     )

I think I've played around with that before at a Hackathon but never got around to finish it. I'll see if I still have something of that lying around.

ckrybus:
Thanks for your feedback, I completely forgot about nested columns. 

I realized that I will need nested columns support too, so I will close this pull request for now and 
should I implement nested columns before you do I will create a new pull request.

mfussenegger:
Okay, would be cool if you can come up with something.

this was the thing I'd been playing around with: https://github.com/crate/crate-python/commit/6a2ea2f6e323c9db36fc155fe84cc21dbf216e4f#diff-c7f6cb482c711062006a50210b16ecc5R25

But I think the change tracking didn't work correctly (which is required to generate correct update statements. E.g that a statement like `update t set obj['x'] = ...`  is generated instead of `update t set obj = {x = ...}`

PR #134, opened 2015-10-09
Fix sqlalchemy create table compiler

ckrybus:
The create table compiler produces incorrect sql when  'crate_clustered_by' and  'crate_number_of_shards'  ``__table_args__`` are used simultaneously.

In this example I'm using a minimal SQL generated by sqlalchemy:
```sql
cr> CREATE TABLE test (
        p STRING
) CLUSTERED INTO 3 SHARDS CLUSTERED BY (p);
CREATE OK (0.077 sec)

cr> select table_name, number_of_shards from information_schema.tables where table_name='test11';
+------------+------------------+
| table_name | number_of_shards |
+------------+------------------+
| test       |                5 |
+------------+------------------+
SELECT 1 row in set (0.002 sec)
```
The number of shards is 5, but it should be 3 ...

The syntax doesn't seem to be valid (according to "[ CLUSTERED [ BY (routing_column) ] INTO num_shards SHARDS ]"), but the crate client generates this sql and the crate database doesn't throw a syntax error (both crash and admin console), instead it seems that only the last "CLUSTERED" statement is used and the first is ignored.

### DIFF
--- a/CHANGES.txt
+++ b/CHANGES.txt
@@ -5,6 +5,8 @@ Changes for crate
 Unreleased
 ==========
 
+- fix: use proper CLUSTERED clause syntax in SQLAlchemy's create table statement
+
 2015/08/12 0.13.4
 =================
 
--- a/src/crate/client/sqlalchemy/compiler.py
+++ b/src/crate/client/sqlalchemy/compiler.py
@@ -19,6 +19,9 @@
 # with Crate these terms will supersede the license and you may use the
 # software solely pursuant to the terms of the relevant commercial agreement.
 
+import string
+from collections import defaultdict
+
 import sqlalchemy as sa
 try:
     # SQLAlchemy 0.8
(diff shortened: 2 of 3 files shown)


chaudum:
hi @ckrybus 

thanks for your contribution. looks good to me!
one small thing: could you please also add a changelog entry to [CHANGES.txt](https://github.com/crate/crate-python/blob/master/CHANGES.txt) (underneath `Unreleased`)?
once you've signed a [CLA](https://crate.io/community/contribute/agreements/) I can merge your pull request.

best,
christian

ckrybus:
@chaudum all done :)

chaudum:
perfect, thanks again for the contribution!

chaudum:
@ckrybus  new release crate [0.13.5](https://pypi.python.org/pypi/crate) is available!

PR #138, opened 2015-10-24
remove lovely.testlayers dependency

mfussenegger:
The functionality that was used of the ServerLayer is
basically just subprocess Popen and kill.

Everything else was either un-used (e.g. stdout/stderr
handling) or unneeded (waiting for the process to become
responsive on its port - the crate layer has that
functionality already with the addition of waiting for HTTP
200 status codes)

### DIFF
--- a/src/crate/testing/layer.py
+++ b/src/crate/testing/layer.py
@@ -7,19 +7,21 @@ import logging
 import urllib3
 import tempfile
 import shutil
+import subprocess
 from urllib3.exceptions import MaxRetryError
-from lovely.testlayers import server, layer
 
 logger = logging.getLogger(__name__)
 
-class CrateLayer(server.ServerLayer):
+class CrateLayer(object):
     """
     this layer starts a crate server.
     """
 
+    __bases__ = ()
+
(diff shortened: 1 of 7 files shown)


src/crate/testing/layer.py
```diff
@@ -93,21 +99,37 @@ def create_settings(self,
...
+        self.process = subprocess.Popen(self.start_cmd)
+        returncode = self.process.poll()
+        if returncode is not None:
+            raise SystemError('Failed to start server rc={0} cmd={1}'.format(
+                returncode, self.start_cmd))
+        self._wait_for_start()
+        self._wait_for_master()
+
+    def stop(self):
+        self.process.kill()
```
chaudum: see comment from removed `stop()` method.
is that not true any more?

chaudum:
otherwise ack

src/crate/testing/layer.py
mfussenegger: didn't see any terminal poisoning locally or on travis. So I assumed that it is no longer true

PR #140, opened 2015-11-02
remove wheel and docutils dependency in base.cfg

mfussenegger:
these dependencies are only useful if one wants to do a release and are
generally not required to run tests


chaudum:
ack

PR #145, opened 2015-12-01
add support for datetime serialization

mfussenegger:
Other db-api implementations (e.g. sqlite3) support datetime instances
in `execute` calls.

E.g. it is possible to do something like:

    c.execute('insert into t (dt) values (?)', (datetime.utcnow(),))

This commit adds the same functionality.

### DIFF
--- a/src/crate/client/http.py
+++ b/src/crate/client/http.py
@@ -30,6 +30,7 @@ import six
 import urllib3
 import urllib3.exceptions
 from time import time
+from datetime import datetime
 import threading
 import re
 from six.moves.urllib.parse import urlparse, urlencode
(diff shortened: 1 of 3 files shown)


src/crate/client/http.py
```diff
@@ -67,6 +68,13 @@ def super_len(o):
         return len(o.getvalue())
 
 
+class CrateJsonEncoder(json.JSONEncoder):
+    def default(self, o):
+        if isinstance(o, datetime):
+            return int(o.timestamp() * 1000)
```
chaudum: `timestamp()` method is python3 only :(

chaudum:
ack

PR #152, opened 2015-12-16
added test for bulk_save_objects()

chaudum:
which is new in SQLAlchemy 1.0

see: http://docs.sqlalchemy.org/en/rel_1_0/orm/persistence_techniques.html#bulk-operations


mfussenegger:
ack

PR #156, opened 2016-02-07
couple of clean-ups

mfussenegger:


src/crate/client/http.py
```diff
@@ -118,12 +118,111 @@ def request(self,
         )
 
 
+def _json_loads_or_error(response):
```
chaudum: think the naming is a bit misleading, because `_json_loads` suggests to parse a string, but the method requires a response object

mfussenegger: what about `_json_from_response` ?

chaudum: sounds good!

mfussenegger: updated the PR

chaudum:
travis failure seems to be unrelated
ack

PR #160, opened 2016-03-17
increased retry_interval in  ThreadSafeHttpClientTest to ensure that

bogensberger:
servers which where set to inactive state
will not be put back to active state immediately, which would cause an endless loop


aslanbekirov:
ACK

PR #161, opened 2016-03-18
disable http retries on read errors

seut:
Read retries by default are set to 3, which will send an SQL statement up to 4 times if e.g. an read timeout occurs. This commit disables read retries at all.

### DIFF
--- a/src/crate/client/test_http.py
+++ b/src/crate/client/test_http.py
@@ -21,6 +21,7 @@
 
 import json
 import time
+import multiprocessing
 import sys
 import os
 from .compat import queue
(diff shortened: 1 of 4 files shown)


src/crate/client/test_http.py
```diff
@@ -150,26 +150,26 @@ def test_redirect_handling(self):
             # 4201 gets added to serverpool but isn't available
             pass
         self.assertEqual(
-            ['http://localhost:4200', 'http://localhost:4201'],
-            sorted(list(client.server_pool.keys()))
+                ['http://localhost:4200', 'http://localhost:4201'],
```
mfussenegger: why the formatting changes? It was valid PEP8 before

src/crate/client/test_http.py
```diff
@@ -362,10 +360,59 @@ def test_no_params(self):
...
+    """
+    http handler for use with BaseHTTPServer
+
+    counts request made
+    """
+
+    def do_GET(self):
+        self.server.SHARED['count'] += 1
+
+    do_POST = do_PUT = do_DELETE = do_HEAD = do_GET
```
mfussenegger: `client.sql` only does POST requests, why do we need the other do_ things?

src/crate/client/test_http.py
```diff
@@ -362,10 +360,59 @@ def test_no_params(self):
...
+
+class RetryOnTimeoutServerTest(TestCase):
+    server_address = ("127.0.0.1", 65535)
+
+    def __init__(self, *args, **kwargs):
+        super(RetryOnTimeoutServerTest, self).__init__(*args, **kwargs)
+        self.server_process = Process(target=TestingHTTPServer.run_server, args=(self.server_address,))
+
+    def setUp(self):
+        super(RetryOnTimeoutServerTest, self).setUp()
```
mfussenegger: it is not really required to call setUp or tearDown of TestCase, the default implementations do nothing

mfussenegger:
ack

PR #162, opened 2016-04-06
Support HTTP Basic Auth

terwey:
Implements HTTP basic auth for connecting to a Crate.io server behind e.g. a haproxy with basic auth

### DIFF
--- a/src/crate/client/http.py
+++ b/src/crate/client/http.py
@@ -35,6 +35,8 @@ from datetime import datetime, date
 import calendar
 import threading
 import re
+from base64 import b64encode
+from string import split
 from six.moves.urllib.parse import urlparse
 from crate.client.exceptions import (
     ConnectionError,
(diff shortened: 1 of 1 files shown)


terwey:
Not sure if the HTTPS already failed before or not. The `string.split()` was indeed a bug from my side. Let me know once this is merged to upstream so we can continue using Crash from upstream.

src/crate/client/http.py
```diff
@@ -87,6 +88,12 @@ def default(self, o):
 class Server(object):
 
     def __init__(self, server, **kwargs):
+        self.auth = None
+        if "@" in server:
+            auth = server.split("@")
+            if "//" in auth[0]:
+                auth = auth[0].split("//")[1]
+            self.auth = auth.split(":")
```
mfussenegger: wouldn't it make sense to already prepare the complete header entry here? Otherwise some part of the work is done here and the rest is done on each request.

mfussenegger:
Sorry for the long delay. Didn't notice this earlier.

I've left a comment. Some tests would also be nice. And before we can merge it
you'd also have to sign our
[CLA](https://crate.io/community/contribute/agreements)

chaudum:
As per 2.3.0 CrateDB supports password authentication and for HTTP is is implemented using Basic Auth. See https://github.com/crate/crate-python/pull/247

PR #164, opened 2016-05-13
Move cert removal for non-https into a separate method

mfussenegger:
Also adds RequestsCaBundleTest to the testsuite. Before it wasn't
executed


chaudum:
ack

PR #167, opened 2016-05-17
Drop support for Python 2.6

mfussenegger:
Python 2.6 reached end of line in 2013 and since then doesn't receive
any security updates.


chaudum:
ack

chaudum:
can you also add an impediment that all references on the website should be updated once released

PR #171, opened 2016-06-13
close connections cleanly in tests

chaudum:


seut:
lgtm

src/crate/client/tests.py
```diff
@@ -250,8 +247,14 @@ def tearDownWithCrateLayer(test):
     # clear testing data
     conn = connect(crate_host)
```
mfussenegger: you could use with ``connect(crate_host) as conn`` to avoid the try/except stuff below

src/crate/testing/layer.py
```diff
@@ -196,50 +209,65 @@ def wdPath(self):
...
+        else:
+            sys.stderr.write('HTTP: {}\n'.format(self.http_url))
+            self._wait_for_start()
+            self._wait_for_master()
+            sys.stderr.write('\nCrate instance ready.\n')
 
     def stop(self):
-        self.process.kill()
+        if self.process:
+            self.process.terminate()
```
mfussenegger: why terminate instead of kill?

mfussenegger: nevermind, terminate is better than kill =)

mfussenegger:
lgtm

PR #179, opened 2016-08-05
Remove certificate keywords when adding non-https server

SzeMengTan:
This pull request addresses the following exception which can occur while fetching blob data from a cluster with multiple non-https servers:

```
  File "BlobAccess.py", line 29, in <module>
    iter = blob_container.get(blob_dict[k])
  File "C:\Anaconda2\envs\sandbox\lib\site-packages\crate\client\blob.py", line 80, in get
    return self.conn.client.blob_get(self.container_name, digest)
  File "C:\Anaconda2\envs\sandbox\lib\site-packages\crate\client\http.py", line 347, in blob_get
    response = self._request('GET', _blob_path(table, digest), stream=True)
  File "C:\Anaconda2\envs\sandbox\lib\site-packages\crate\client\http.py", line 406, in _request
    raise ProgrammingError(_ex_to_message(e))
crate.client.exceptions.ProgrammingError: __init__() got an unexpected keyword argument 'cert_file'
```

### DIFF
--- a/src/crate/client/http.py
+++ b/src/crate/client/http.py
@@ -365,7 +365,8 @@ class Client(object):
     def _add_server(self, server):
         with self._lock:
             if server not in self.server_pool:
-                self.server_pool[server] = Server(server, **self._pool_kw)
+                kwargs = _remove_certs_for_non_https(server, self._pool_kw)
+                self.server_pool[server] = Server(server, **kwargs)
 
     def _request(self, method, path, server=None, **kwargs):
         """Execute a request to the cluster
--- a/src/crate/client/http.py
+++ b/src/crate/client/http.py
@@ -283,10 +283,13 @@ class Client(object):
         for server in self.server_pool.values():
             server.close()
 
+    def _create_server(self, server, **kwargs):
+        kwargs = _remove_certs_for_non_https(server, kwargs)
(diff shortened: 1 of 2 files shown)


src/crate/client/http.py
```diff
@@ -365,7 +365,8 @@ def blob_exists(self, table, digest):
     def _add_server(self, server):
         with self._lock:
             if server not in self.server_pool:
-                self.server_pool[server] = Server(server, **self._pool_kw)
+                kwargs = _remove_certs_for_non_https(server, self._pool_kw)
```
chaudum: This is the same procedure as in line 288-289.
You could extract these two lines into a new method and reuse the method.

chaudum:
Hi @SzeMengTan
Thanks for your contribution! Looking good 😎
I've made a comment about a minor improvement. Would be nice if you could address it.
In order to be able to merge your PR you'd need to sign our contributor license agreement that you can find here: https://crate.io/community/contribute/agreements

Also, please add a changelog entry to the [CHANGES.txt](https://github.com/crate/crate-python/blob/master/CHANGES.txt#L7).

Christian

SzeMengTan:
Hi Christian,

I've refactored the code as you suggested, completed the license agreement and added an entry into the change log. Thank you for your consideration of this matter.

Sze

Issue #180, opened 2016-08-10
__init__() got an unexpected keyword argument 'cert_file'


mheppner:
Using python 2.7, crate==0.16.2, and urllib3==1.16, I get an error when trying to retrieve a blob.

The table was created using:
```sql
create blob table testblobs clustered into 2 shards with (number_of_replicas=1)
```

The file was successfully added using:
```python
from crate import client
ctx = client.connect(['box1.local:4200', 'box1.local:4200', 'box3.local:4200'])
blobs = ctx.get_blob_container('testblobs')

with open('/path/sample.tif') as f:
    blobs.put(f, digest='<sha1hash>')
```

When trying to retrieve:
```python
from crate import client
ctx = client.connect(['box1.local:4200', 'box1.local:4200', 'box3.local:4200'])
blobs = ctx.get_blob_container('testblobs')

res = blobs.get('<sha1hash>')
```
```python
  (...)
  File "/root/test/lib/python2.7/site-packages/crate/client/blob.py", line 80, in get
    return self.conn.client.blob_get(self.container_name, digest)
  File "/root/test/lib/python2.7/site-packages/crate/client/http.py", line 347, in blob_get
    response = self._request('GET', _blob_path(table, digest), stream=True)
  File "/root/test/lib/python2.7/site-packages/crate/client/http.py", line 406, in _request
    raise ProgrammingError(_ex_to_message(e))
crate.client.exceptions.ProgrammingError: __init__() got an unexpected keyword argument 'cert_file'
```

chaudum:
duplicate of #179

PR #181, opened 2016-08-12
added regression test for adding non-https servers on redirect

chaudum:
regression test for #179 / #180 

@mfussenegger wanna take a look?


bogensberger:
ack

PR #184, opened 2016-10-17
Fix sqlalchemy: updates in nested object columns

bogensberger:
have been ignored

### DIFF
--- a/src/crate/client/sqlalchemy/types.py
+++ b/src/crate/client/sqlalchemy/types.py
@@ -90,22 +90,44 @@ class MutableDict(Mutable, dict):
         else:
             return value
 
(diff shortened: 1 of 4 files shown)


src/crate/client/sqlalchemy/types.py
```diff
@@ -90,23 +90,45 @@ def coerce(cls, key, value):
         else:
             return value
 
-    def __init__(self, initval=None):
+    def __init__(self, initval=None, to_update=None, root_change_key=None):
```
kovrus: maybe **kargs, if it makes sense of course

kovrus: otherwise ack

Issue #187, opened 2016-11-16
sqlalchemy update in python 3.5


Negashev:
### I have model `User`
```python
# __init__.py
from flask_sqlalchemy import SQLAlchemy

from server import app

db = SQLAlchemy(app)

from server.db.models import User

db.create_all()
# models.py
class User(db.Model):
    __tablename__ = 'users'
    mail = db.Column(db.String, primary_key=True)
    img = db.Column(db.String)

    def __init__(self, mail):
        self.mail = mail
        self.img = hashlib.md5(mail.encode()).hexdigest()

    def __repr__(self):
        return '<User %r>' % self.mail

```

### When i try to update on row, i get error
```python
>>> import server.db as h
>>> s=h.models.User.query.filter_by(mail='i@negash.ru').first()
>>> s.img='965fe6f7e5727271c4c67fc43f325288_NEW_DATA'
>>> h.db.session.commit()
```

```python
Traceback (most recent call last):
  File "<stdin>", line 1, in <module>
  File "/usr/local/lib/python3.5/site-packages/sqlalchemy/orm/scoping.py", line 157, in do
    return getattr(self.registry(), name)(*args, **kwargs)
  File "/usr/local/lib/python3.5/site-packages/sqlalchemy/orm/session.py", line 874, in commit
    self.transaction.commit()
  File "/usr/local/lib/python3.5/site-packages/sqlalchemy/orm/session.py", line 461, in commit
    self._prepare_impl()
  File "/usr/local/lib/python3.5/site-packages/sqlalchemy/orm/session.py", line 441, in _prepare_impl
    self.session.flush()
  File "/usr/local/lib/python3.5/site-packages/sqlalchemy/orm/session.py", line 2137, in flush
    self._flush(objects)
  File "/usr/local/lib/python3.5/site-packages/sqlalchemy/orm/session.py", line 2257, in _flush
    transaction.rollback(_capture_exception=True)
  File "/usr/local/lib/python3.5/site-packages/sqlalchemy/util/langhelpers.py", line 60, in __exit__
    compat.reraise(exc_type, exc_value, exc_tb)
  File "/usr/local/lib/python3.5/site-packages/sqlalchemy/util/compat.py", line 186, in reraise
    raise value
  File "/usr/local/lib/python3.5/site-packages/sqlalchemy/orm/session.py", line 2221, in _flush
    flush_context.execute()
[...]

PR #188, opened 2016-11-18
added compatibility to sqlalchemy version 1.1

mikethebeer:
fixes https://github.com/crate/crate-python/issues/187

@chaudum pls take a 👀


src/crate/client/sqlalchemy/sa_version.py
```diff
@@ -25,6 +25,10 @@
 def is_sa_version(major, minor):
     return int(SA_VERSION[0]) == major and int(SA_VERSION[1]) == minor
 
+def gte_sa_version(major, minor):
+    return int(SA_VERSION[0]) == major and int(SA_VERSION[1]) >= minor
```
mfussenegger: btw it is possible to do tuple comparisons:

```
>>> (1, 0) >= (1, 1)
False
>>> (1, 1) >= (1, 1)
True
>>> (1, 2) >= (1, 1)
True
>>> (2, 0) >= (1, 1)
True
>>> (0, 9) >= (1, 1)
False
```

versions.cfg
```diff
@@ -9,7 +9,7 @@ Jinja2 = 2.7.3
 MarkupSafe = 0.23
 Pygments = 1.6
 Sphinx = 1.2.3
-SQLAlchemy = 1.0.10
+SQLAlchemy = 1.0.16
```
chaudum: you could even update to `1.1.4`

PR #193, opened 2016-12-06
testing: updated CrateDB version to 1.0.0

chaudum:
@seut please

### DIFF
--- a/src/crate/client/http.txt
+++ b/src/crate/client/http.txt
@@ -96,10 +96,10 @@ Check if a blob exists::
 
 Trying to get a non-existing blob throws an exception::
 
-    >>> http_client.blob_get('myfiles', '040f06fd774092478d450774f5ba30c5da78acc8')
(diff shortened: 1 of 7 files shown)


src/crate/client/http.txt
```diff
@@ -153,7 +153,7 @@ Uploading a blob to a table with disabled blob support throws an exception::
     ...     'locations', '040f06fd774092478d450774f5ba30c5da78acc8', f)
     Traceback (most recent call last):
     ...
-    BlobsDisabledException: locations/040f06fd774092478d450774f5ba30c5da78acc8
+    ProgrammingError: 404 Client Error: Not Found
```
chaudum: @mfussenegger is this change intendend?

mfussenegger: @chaudum looked into it
We changed the error handling to return 404 instead of 400 if a index/shard is not found.
I think we should change https://github.com/crate/crate-python/blob/master/src/crate/client/http.py#L330

chaudum: @mfussenegger I've pushed a fixup

PR #199, opened 2016-12-15
B/drop legacy sqlalchemy

bogensberger:


mfussenegger:
Maybe we should have a compatibility matrix in the readme or somewhere? /cc: @joemoe

PR #201, opened 2017-01-19
Remove multicast option from the test layer as it will no longer be available for next Crate versions

matriv:


seut:
pls add a changes entry prefixed with "Testing:"

PR #203, opened 2017-01-26
Fix cmd line settings to be passed with -C and not -Des.

matriv:
which is deprecated and will be removed in newer crate versions.


mfussenegger:
Please add a entry in the CHANGES.txt. This breaks the test layer for use with older crate versions

PR #204, opened 2017-01-31
Convert boolean True/False values for additional settings

matriv:
of testing layer to lowercase strings 'true'/'false'.

### DIFF
--- a/src/crate/testing/layer.py
+++ b/src/crate/testing/layer.py
@@ -201,8 +201,9 @@ class CrateLayer(object):
                                         port or '4200-4299',
                                         transport_port or '4300-4399',
                                         settings)
-        start_cmd = (crate_exec, ) + tuple(["-C%s=%s" % opt
-                                            for opt in settings.items()])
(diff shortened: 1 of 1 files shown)


src/crate/testing/layer.py
```diff
@@ -201,8 +201,9 @@ def __init__(self,
                                         port or '4200-4299',
                                         transport_port or '4300-4399',
                                         settings)
-        start_cmd = (crate_exec, ) + tuple(["-C%s=%s" % opt
-                                            for opt in settings.items()])
+        # ES 5 cannot parse 'True'/'False' as booleans so convert to lowercase
+        start_cmd = (crate_exec, ) + tuple(["-C%s=%s" % ((key, str(value).lower()) if type(value) == bool else (key, value))
```
mfussenegger: `"%s"` already does `str(value)` so I think the same could also be written as `["-C%s=" % key + str(value).lower() for key, value in settings.item()]` without the explicit bool check

PR #207, opened 2017-02-16
added support for env. variables to testing layer

kovrus:


src/crate/testing/layer.py
```diff
@@ -155,7 +155,8 @@ def __init__(self,
                  cluster_name=None,
                  host="127.0.0.1",
                  settings=None,
-                 verbose=False):
+                 verbose=False,
+                 env={}):
```
mfussenegger: Please use `env=None`. Default values are bound at method definition time, not at evaluation time.

```
>>> def foo(x={}):
...     x[random.randint(0, 10)] = 1
...     print(x)
... 
>>> foo()
{4: 1}
>>> foo()
{4: 1, 0: 1}
>>> foo()
{4: 1, 0: 1, 7: 1}
```

src/crate/testing/layer.py
```diff
@@ -182,6 +184,8 @@ def __init__(self,
 
         self.process = None
         self.verbose = verbose
+        self.env = env
+        self.env.update(dict(CRATE_USE_IPV4='true'))
```
mfussenegger: Either use `self.env['CRATE_USE_IPV4'] = 'true'` because it's only setting a single value, or better `self.env.setdefault` to allow a user to explicitly override the default.

PR #210, opened 2017-02-23
Spawn a thread to consume the stdout

andreidan:
The stdout pipe buffer is limited (and the size varies across systems). We try to read the crate http url from the startup logs (which is rather weird, shouldn't we be using the default port?) and that is our only consumer for the stdout. 

We encountered a problem with this when we tried to switch crate logging to debug. As considerably more log statements (than the default WARNING level we currently use in tests) are issued the pipe buffer overflows. 

As a solution to this, we spawn a thread that consumes the pipe buffer.

### DIFF
--- a/CHANGES.txt
+++ b/CHANGES.txt
@@ -5,6 +5,9 @@ Changes for crate
 Unreleased
 ==========
 
+ - Testing: Prevent the process.stdout buffer from filling up in the test layer 
+   which in turn would cause the process to block 
+
  - Raise more meaningful `BlobLocationNotFoundException` error when
    trying to upload a file to an invalid blob table.
 
--- a/src/crate/testing/layer.py
+++ b/src/crate/testing/layer.py
@@ -30,6 +30,7 @@ import shutil
 import subprocess
 import tarfile
 import io
+import threading
 
 try:
     from urllib.request import urlopen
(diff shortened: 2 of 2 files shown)


src/crate/testing/layer.py
```diff
@@ -277,11 +284,25 @@ def start(self):
...
             sys.stderr.write('\nCrate instance ready.\n')
 
+    def out_thread_stopped(self):
+            return self._stop_out_thread.isSet()
+
+    def stop_out_thread(self):
+            self._stop_out_thread.set()
+
+    def consume(self, iterable):
+        while not self.out_thread_stopped():
```
mfussenegger: I think since the iterable is infinite/blocking (as it's linked to stdout of the subprocess) this condition will never be hit?

mfussenegger:
Please add to the commit message a short sentence explaining the motivation behind this change.

src/crate/testing/layer.py
```diff
@@ -269,6 +272,10 @@ def start(self):
             # this is necessary if no static port is assigned
             self.http_url = wait_for_http_url(self.process.stdout, verbose=self.verbose)
 
+        self._out_thread = threading.Thread(target=self._consume, args=(self.process.stdout,))
```
mfussenegger: Could also add a comment here that this is to unblock the process.stdout buffer from filling up, which would cause the process to block

mfussenegger:
Oh, and I think a CHANGES.txt entry is necessary

andreidan:
retest this please

CHANGES.txt
```diff
@@ -5,6 +5,9 @@ Changes for crate
 Unreleased
 ==========
 
+ - Prevent the process.stdout buffer from filling up which in turn would cause 
+   the process to block
```
mfussenegger: Please make it clear that this is test-layer related

PR #211, opened 2017-02-27
Prepare release 0.19.1

andreidan:
Release changes: 

 - Testing: Prevent the process.stdout buffer from filling up in the test layer
   which in turn would cause the process to block
 - Raise more meaningful `BlobLocationNotFoundException` error when
   trying to upload a file to an invalid blob table.

### DIFF
--- a/CHANGES.txt
+++ b/CHANGES.txt
@@ -5,6 +5,9 @@ Changes for crate
 Unreleased
 ==========
 
+2017/02/27 0.19.1
+=================
+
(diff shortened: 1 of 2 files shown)


CHANGES.txt
```diff
@@ -2,8 +2,8 @@
 Changes for crate
 =================
 
-Unreleased
```
kovrus: do not remove the `Unreleased` section

PR #212, opened 2017-03-06
improve top level docs

nslater:
the best way to preview these changes is to browse them in GitHub directly

https://github.com/crate/crate-python/tree/nomi/top-level-docs


DEVELOP.rst
```diff
@@ -1,84 +1,94 @@
...
-    ./bin/tox -e py33
+To run against a single interpreter, you can also do::
 
-Note: Before running tests make sure to stop all crate instances which
-transport port is listening on port 9300 to avoid side effects with the test
-layer.
+    $ ./bin/tox -e py33
 
+*Note*: before running tests make sure to stop all CrateDB instances which
+listening on port 9300 to avoid side effects with the test layer.
```
kovrus: we do use port 9300 as default for anything. We can say that a user has to assure that there are no locally running instances which listen on the default crate transport port 4300.

nslater:
@kovrus could you review my new commits. if you're happy I'll squash and rebase

PR #216, opened 2017-03-29
improve developer docs

nslater:


DEVELOP.rst
```diff
@@ -66,6 +66,23 @@ Then, use twine_ to upload the packages::
 
     $ bin/twine upload dist/*
 
+This last command builds the package and uploads it to PyPI_.
```
chaudum: `bin/twine` only uploads, does not build

nslater:
@chaudum good catch. updated!

PR #224, opened 2017-05-17
fix optimization for bulk updates

mxm:
- don't optimize if parameters are not supplied yet
- check the key type before binding parameters


mxm:
Thanks for the comments @mfussenegger. I've added an integration test and updated the CHANGES file.

CHANGES.txt
```diff
@@ -5,6 +5,13 @@ Changes for crate
 Unreleased
 ==========
 
+2017/05/17 0.19.3
+=================
+
+ - Fix bulk updates which were broken due to query rewrites.
+ - Fix typo to DB API in README
```
mfussenegger: There's no need to mention the README fix - only end-user relevant changes should end up in the changes.txt

jeffnappi:
Thanks for jumping in on this @mxm - I was just about to jump in and fix it myself :)

Looks like the Travis Tests are failing simply due to the missing newline in `update_test.py`

Testing now to see if this resolves my issue - appears like it should.

CHANGES.txt
mxm: Makes sense. Thanks.

mxm:
Thanks @jeffnappi. I fixed the missing newline. Let's wait for the tests to pass.

mxm:
One of my new tests is failing with SQLAlchemy 1.0.16 but passes with 1.1.4. Let's see.

mxm:
@mfussenegger Is this good to merge?

PR #227, opened 2017-06-21
docs fixes

dornbirndevelops:


docs/conf.py
```diff
@@ -1 +1,2 @@
 from crate.theme.rtd.conf.crate_python import *
+html_favicon = None
```
jodok: you should always add a newline at the end of the file :) (you can also configure your editor to do this automatically)

chaudum: @jodok there is already a newline character at the end of the line
python should not have a blank line at the end of the file (pep8 W391)

```console
$ pep8 docs/conf.py
docs/conf.py:4:1: W391 blank line at end of file
```

jodok: what does that symbol mean? 
image

chaudum: @jodok is right, the newline character is missing

dornbirndevelops: I'm sry that rule was new to me. I'll fix it.

PR #228, opened 2017-06-22
Enforce cert check when verify_ssl_cert=True

mxm:
The Client class simply skipped the CA certificate check when no
certificate had been supplied but verify_ssl_cert is set to True. From
the outside this looks like the certificate check passed. If no
certificate has been supplied via parameter or environment variable, we
should fail.


mxm:
Of course 👍

PR #229, opened 2017-06-23
clarify release process in DEVELOP

mxm:
I wanted to clarify the release process because I released out of master the last time and wasn't aware that there were feature branches. 

Further, I wanted to explicitly state that the release commit also goes to master. IMHO it should only be in the feature branch but it should be fine since we release out of master most of the time.

### DIFF
--- a/DEVELOP.rst
+++ b/DEVELOP.rst
@@ -45,15 +45,27 @@ Preparing a Release
 
 To create a new release, you must:
 
+- Backport your bug fixes to the latest stable branch x.y (e.g. 0.19)
+
+- For new features, create a new stable branch x.(y+1) (e.g. 0.20)
+
+In the release branch:
+
 - Update ``__version__`` in ``src/crate/client/__init__.py``
 
 - Add a section for the new version in the ``CHANGES.txt`` file
 
 - Commit your changes with a message like "prepare release x.y.z"
 
-- Push to origin
+- Push to origin/<release_branch>
 
 - Create a tag by running ``./devtools/create_tag.sh``
+
+On master:
+
+- Update the release notes to reflect the release
+
+Next:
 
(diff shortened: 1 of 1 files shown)


mxm:
The PR just clarifies the current process. In the future we could introduce a master version like the latest release x.y.z with a dev suffix: `x.y.z-dev` or just `x.y`. Then the release commit should only be in the release branch/tag.

DEVELOP.rst
```diff
@@ -45,16 +45,24 @@ Preparing a Release
...
 - Add a section for the new version in the ``CHANGES.txt`` file
 
 - Commit your changes with a message like "prepare release x.y.z"
 
-- Push to origin
+- Push to origin/<release_branch>
 
 - Create a tag by running ``./devtools/create_tag.sh``
 
+On the master branch if the release was based on the master branch:
```
mfussenegger: I'd shorten this to:

    On master:

      - Update the release notes to reflect the release.


If this is done via cherry-pick, merge or even manually doesn't matter. The
result is what's important.

DEVELOP.rst
```diff
@@ -45,16 +45,24 @@ Preparing a Release
 
 To create a new release, you must:
 
+- Backport your changes to the release branch x.y
```
mfussenegger: I think "release branch" is missleading. We've a stable branch and that should only receive bug-fixes. Features are in master. And if a new feature is released it's necessary to create a new stable branch.

DEVELOP.rst
mxm: Good point. Will incorporate that.

DEVELOP.rst
mxm: Thanks, I'll try to make that clearer.

PR #232, opened 2017-06-26
Reflect release 0.20.0

matriv:


src/crate/client/__init__.py
```diff
@@ -24,7 +24,7 @@
 
 # version string read from setup.py using a regex. Take care not to break the
 # regex!
-__version__ = "0.19.5"
+__version__ = "0.20.0"
```
chaudum: I know it has not been like that before, but shouldn't it be `0.21-dev` on master until 0.21 is released? @mfussenegger what do you think?

mfussenegger: Yep I guess that would make sense.

We could also consider to use https://pypi.python.org/pypi/setuptools_scm

PR #237, opened 2017-07-28
Fix deprecation warnings

mxm:
The new Crate version uses ES 5.4.3 which logs these warnings for every REST request:

```
[WARN ][o.e.d.r.RestController   ] Content type detection for rest requests is deprecated. Specify the content type using the [Content-Type] header.
[WARN ][o.e.d.r.RestRequest      ] Expected a boolean [true/false] for request parameter [error_trace] but got [1]
```

### DIFF
--- a/src/crate/client/http.py
+++ b/src/crate/client/http.py
@@ -104,7 +104,7 @@ class Server(object):
                 **kwargs):
         """Send a request
 
-        Always set the Content-Length header.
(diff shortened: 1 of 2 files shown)


mxm:
Fixed the test case. Seems like the tests are flaky for certain versions. Test failures now seem to be unrelated.

PR #243, opened 2017-09-21
Updated tests to use CrateDb 2.1.7.

christianbader:


CHANGES.txt
```diff
@@ -5,6 +5,8 @@ Changes for crate
 Unreleased
 ==========
 
+ - Updated tests to use CrateDb 2.1.7.
```
chaudum: CrateDB

chaudum: tests do not require a changelog entry

src/crate/testing/layer.py
```diff
@@ -44,7 +44,7 @@
 
 CRATE_CONFIG_ERROR = 'crate_config must point to a folder or to a file named "crate.yml"'
 HTTP_ADDRESS_RE = re.compile(
-    '.*\[http +\] \[.*\] .*'
+    '.*\[i.c.p.h.CrateNettyHttpServerTransport] \[.*\] .*'
```
mfussenegger: Should we keep compat here for the older format? If not it's kind of a breaking changes for the testing layer.

christianbader: Yes, good point

chaudum: I agree, we should keep it compatible.

PR #244, opened 2017-10-26
Prepare Python client for CrateDB 2.3.0

mxm:
With https://github.com/crate/crate/pull/6448, we have removed the pk column information from `table_constraints` and populate the `key_column_usage` table correctly. The Python client needs to be updated to reflects theses changes. We have to maintain backwards-compatibility for older CrateDB versions.


mxm:
Seems like the tests on Travis are failing when retrieving dependencies. Locally the tests run fine...

src/crate/client/sqlalchemy/tests/dialect_test.py
```diff
@@ -0,0 +1,71 @@
...
+        insp = inspect(meta.bind)
+        self.engine.dialect.server_version_info = (0, 54, 0)
+        fake_cursor.rowcount = 1
+        fake_cursor.fetchone = MagicMock(return_value=[["id", "id2", "id3"]])
+        eq_(insp.get_pk_constraint("characters")['constrained_columns'], ["id", "id2", "id3"])
+        # test the new pk retrieval
+        insp = inspect(meta.bind)
+        self.engine.dialect.server_version_info = (2, 3, 0)
+        fake_cursor.rowcount = 3
+        fake_cursor.fetchall = MagicMock(return_value=[["id"], ["id2"], ["id3"]])
```
mfussenegger: would it make sense to also verify the statement that was executed? otherwise it looks as if this test verifies that the mock works

mxm: This is what the invariants checks below do: 

```java
fake_cursor.fetchone.assert_called_once_with()
fake_cursor.fetchall.assert_called_once_with()
```

mxm: I moved them directly behind the `_eq` check.

mxm: You mean the SQL statement?

mxm:
Seems like the added test and all other sqlalchemy tests are not run by `bin/test`. They can only be run manually. Will have to look into that before merging...

mxm:
@mfussenegger Addressed the test execution issue and added a check for the actual query executed.

PR #247, opened 2017-12-07
Added password parameter for DB authentication

matriv:


src/crate/client/http.py
```diff
@@ -112,8 +114,11 @@ def request(self,
             length = super_len(data)
             if length is not None:
                 headers['Content-Length'] = length
-        if 'X-User' not in headers and username is not None:
-            headers['X-User'] = username
+        if 'Authorization' not in headers and username is not None:
```
chaudum: you could use the `make_headers()` helper from urllib3 https://github.com/shazow/urllib3/blob/release/urllib3/util/request.py#L11

matriv: make_headers() creates a new dict object that I have to merge with the headers dict we already have. Since the b64encode is not that complex code, I'd prefer to keep it like it is.

mfussenegger: Can we come up with a solution here that is still backward compatible? E.g. if someone is using a username right now with crate 2.2. and then he upgrades it will break because crate 2.2. doesn't know the Authorization header

mxm: Good point. We definitely need to do a follow-up for this. We should only send the new header to servers with version >= 2.3.0

PR #249, opened 2017-12-07
Send username also with 'X-User' to keep backwards compatibility.

matriv:
Followup of b260d36aa7a03e89a737450d8560fc39af260c2c.

### DIFF
--- a/src/crate/client/http.py
+++ b/src/crate/client/http.py
@@ -114,11 +114,18 @@ class Server(object):
             length = super_len(data)
             if length is not None:
                 headers['Content-Length'] = length
-        if 'Authorization' not in headers and username is not None:
-            credentials = username + ':'
-            if (password is not None):
-                credentials += password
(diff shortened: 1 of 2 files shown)


src/crate/client/http.py
```diff
@@ -114,11 +114,17 @@ def request(self,
...
-            headers['Authorization'] = 'Basic %s' % b64encode(credentials.encode('utf-8')).decode('utf-8')
+
+        # Authentication credentials
+        if username is not None:
+            if 'Authorization' not in headers and username is not None:
+                credentials = username + ':'
+                if (password is not None):
+                    credentials += password
+                headers['Authorization'] = 'Basic %s' % b64encode(credentials.encode('utf-8')).decode('utf-8')
+            if 'X-User' not in headers:
```
mfussenegger: maybe add a comment that this is for backward compatibility with cratedb <= 2.2

PR #257, opened 2018-01-09
Drop support for Python 2.7 and 3.3

chaudum:


README.rst
```diff
@@ -38,9 +38,9 @@ This library:
 Prerequisites
 =============
 
-Recent versions of this library require Python (>= 2.7) to run.
+The latest versions of this library requires **Python 3** (>= 3.4).
 
-Use library version 0.14 if you're running Python 2.6.
+Use version ``0.21.x`` for Python 2.7/3.3 and ``0.14`` for Python 2.6.
```
mfussenegger: Would it make sense that recent versions of pip default to installing `0.21.x` if using python2.7/3.3 ?

src/crate/client/exceptions.py
```diff
@@ -19,21 +19,19 @@
...
 
-from .compat import StandardError
 
-
-class Error(StandardError):
+class Error(Exception):
 
     def __init__(self, msg=None, error_trace=None):
-        # for py33 compat.
+        # for py3 compat.
```
mfussenegger: Might make sense to rephrase the comment here - given that we only support py3 it's kinda strange to point out that this is for py3 compatibility.

I guess now we want this to not break clients already using the message attribute?

PR #260, opened 2018-01-10
testing: fixed bin/tox runner

chaudum:


src/crate/client/http.txt
```diff
@@ -158,11 +158,25 @@ Uploading a blob to a table with disabled blob support throws an exception::
...
+    >>> http_client.sql('''
+    ... CREATE FUNCTION fib(LONG) RETURNS LONG
+    ... LANGUAGE JAVASCRIPT AS '
+    ...   var fib = function fib(n) { return n < 2 ? n : fib(n-1) + fib(n-2); }
+    ... '
+    ... ''')
+    {'cols': [], 'rows': [[]], 'rowcount': 1, 'duration': ...}
+
+It's possible to define a HTTP timeout in seconds on client instantiation, so a
+timeout exception is raised when timeout is reached::
```
matriv: I would say `,so an exception is raised when the timeout is reached`

PR #264, opened 2018-01-15
replace underscore with hypen in URL

nslater:


m-kharbat:
@nslater tests are failing with `FileNotFoundError: [Errno 2] No such file or directory: '/home/travis/build/crate/crate-python/src/crate/client/../../../docs/advanced_usage.txt'
 `

PR #267, opened 2018-01-16
testing: reduced timeout in client test

chaudum:
sometimes the test query took less than 100ms


matriv:
@chaudum plz update the commit message to mention the timeout arguement refactoring

PR #269, opened 2018-01-26
Set watermark.flood_stage to avoid config parsing failure

mxm:
In ES 6 you can either specify a unit (e.g. bytes) or a percentage of the disk
size. If you mix the two, you'll get an error.

### DIFF
--- a/CHANGES.txt
+++ b/CHANGES.txt
@@ -9,6 +9,11 @@ Unreleased
    If you are using crate with Python 2.7 or 3.3 already, you will not be able
    to install newer versions of this package.
 
+ - BREAKING: In the testing layer, the custom setting of
+ `cluster.routing.allocation.disk.watermark.low` (1b) and
+ `cluster.routing.allocation.disk.watermark.high` (1b) has been removed.
+ These now default to 85% and 90%, respectively.
+
 2018/01/03 0.21.1
 =================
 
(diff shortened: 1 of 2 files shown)


mxm:
This is a new settings, that's why tests fail. https://www.elastic.co/guide/en/elasticsearch/reference/6.x/disk-allocator.html

chaudum:
I think we should try to make the testing layer backwards compatible.

mxm:
Absolutely. I would opt for using percentages and use the defaults (85/95%) or set them higher. I don't see how the current setting of '1b' works other than that ES seems to ignore this low limit.

The benefit of using percentages is that the default for flood_stage is in percentage (95%) and we can skip its definition then, making the config backwards-compatible.

Do you think that would work (see fixups).

chaudum:
I think your fixup is the best solution, which leaves the default values.
The idea of the `1b` value was that you would never run into the low/high disk watermark which of course could lead to problems during testing.
However, I prefer to set these setting in the actual usage of the layer, not in the testing layer itself - to keep backwards compatibility of the layer.
@mfussenegger what do you think?

CHANGES.txt
```diff
@@ -9,6 +9,11 @@ Unreleased
    If you are using crate with Python 2.7 or 3.3 already, you will not be able
    to install newer versions of this package.
 
+ - BREAKING: In the testing layer, the custom setting of
```
chaudum: The initial change was to add the flood_stage watermark setting, which was a breaking changes. However, in the end, we decided to remove the other settings, which is not a breaking change.

mxm: I get what you mean. Still, if you rely on these default settings, it is breaking and not backwards-compatible.

PR #284, opened 2018-03-25
 pin setuptools version in travis virtualenv

chaudum:


.travis.yml
```diff
@@ -14,7 +14,9 @@ addons:
 before_install:
   # clean existing dependencies to avoid clashes
   - pip freeze | xargs pip uninstall -y
-  - pip install --upgrade setuptools
+  - pip uninstall -y setuptools
+  - pip install setuptools==33.1.1  # pin specific version in virtualenv
```
mxm: Why freeze? Wouldn't it work to only uninstall? Buildout should install a new version.

PR #287, opened 2018-04-10
added schema keyword argument to connect() method

chaudum:
The client now allows to define a different default schema when connecting to
CrateDB with the ``schema`` keyword argument. This causes all statements and
queries that do not specify a schema explicitly to use the provided schema.

### DIFF
--- a/src/crate/client/test_http.py
+++ b/src/crate/client/test_http.py
@@ -21,6 +21,7 @@
 
 import json
 import time
+import socket
 import multiprocessing
 import sys
 import os
@@ -414,7 +415,6 @@ class ParamsTest(TestCase):
 
 class RequestsCaBundleTest(TestCase):
 
-
     def test_open_client(self):
         os.environ["REQUESTS_CA_BUNDLE"] = CA_CERT_PATH
         try:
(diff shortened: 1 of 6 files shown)


src/crate/client/test_http.py
```diff
@@ -436,15 +436,30 @@ def test_remove_certs_for_non_https(self):
...
-class RetryRequestHandler(BaseHTTPRequestHandler):
+class TimeoutRequestHandler(BaseHTTPRequestHandler):
     """
-    http handler for use with HTTPServer
+    HTTP handler for use with TestingHTTPServer
+    updates the shared counter and waits so that the client times out
+    """
+
+    def do_POST(self):
+        print('POST /_sql')
```
mfussenegger: Do we really want to have print statements here?

src/crate/client/test_http.py
```diff
@@ -436,15 +436,30 @@ def test_remove_certs_for_non_https(self):
...
+
+class SharedStateRequestHandler(BaseHTTPRequestHandler):
+    """
+    HTTP handler for use with TestingHTTPServer
+    sets the shared state of the server and returns an empty response
     """
 
     def do_POST(self):
+        print('POST /_sql')
+        print(self.headers)
```
mxm: print left over?

src/crate/client/test_http.py
```diff
@@ -472,79 +495,113 @@ class TestingHTTPServer(HTTPServer):
...
+        super().setUp()
+        self.client = self.clientWithKwargs(schema='my_custom_schema')
+
+    def tearDown(self):
+        super().tearDown()
+        self.client.close()
+
+    def test_default_schema(self):
+        self.client.sql('SELECT 1')
+        self.assertEqual(TestingHTTPServer.SHARED['schema'], 'my_custom_schema')
```
mxm: Should we do an additional test here to see that queries really use the custom schema?

src/crate/client/test_http.py
chaudum: No

src/crate/client/test_http.py
chaudum: That functionality is tested within CrateDB already

mxm: So there is a client test in CrateDB which checks that the client passes on the schema correctly?

chaudum: https://github.com/crate/crate/blob/master/sql/src/test/java/io/crate/integrationtests/RestSQLActionIntegrationTest.java#L71-L83

mxm: Cool!

PR #292, opened 2018-04-24
tidy up .gitignore

nslater:


.gitignore
```diff
@@ -1,16 +1,16 @@
-.coverage
+.coverage/
```
mxm: Is there a semantic difference or did you just want to indicate `.coverage` is a directory?

nslater: yeah just to indicate

PR #296, opened 2018-05-02
use setup.cfg to configure flake8

chaudum:


mfussenegger:
tests are failing?

chaudum:
hm, seems that the config is not picked up.

PR #304, opened 2018-07-30
Don't remove server from active servers if encountering Connection Reset or Broken Pipe errors

autophagy:


autophagy:
@chaudum What do you think? I honestly couldnt find a better way, urllib exposes the underlying exception opaquely. You could always assume the 2nd item in the tuple is the wrapped exception, but I don't know if I'd want to do that.

src/crate/client/http.py
```diff
@@ -425,9 +426,16 @@ def _request(self, method, path, server=None, **kwargs):
                     raise ConnectionError(
                         "Server not available, exception: %s" % ex_message
                     )
-                with self._lock:
-                    # drop server from active ones
-                    self._drop_server(next_server, ex_message)
+                preserve_server = False
```
chaudum: can be written simpler (without the map->list casting)

```py
if isinstance(ex, ProtocolError):
    types = [type(a) for a in ex.args]
    if not any(t in types for t in PRESERVE_ACTIVE_SERVER_EXCEPTIONS):
        with self._lock:
            self._drop_server(...)
```

however, shouldn't it check something else than the `type`,
since `type(ConnectionResetError) == <class 'type'>`

what are typical `type`s of the exception arguments?

autophagy: The typical types of the HTTP/ProtocolException are:
```
[<class 'str'>, <class 'ConnectionResetError'>]
```

Typically the tuple is something like `('Connection aborted.', ConnectionResetError(104, 'Connection reset by peer'))`

You're right that the `type` of `ConnectionResetError` is that of `<class 'type'>`, but thats because you're getting the type of the *class*, not the object. The exception arg's tuple has an instance of the *object*, which has type `<class 'ConnectionResetError'>`

chaudum: @autophagy ok, got it.
However, I would still propose the change from `list(map(...))` to `[type(a) for a in ex.args]` which is more pythonic

autophagy: Totally - i was overcomplicating it. Thank you for the help, will push a fixie :slightly_smiling_face:

chaudum: Also, any chance to get a test case for that?

autophagy: Sure thing

PR #308, opened 2018-11-09
LICENSE docs for docs dir

nslater:


nslater:
@chaudum merged. thanks for spotting!

PR #324, opened 2019-07-12
Add type compat for 4.0 and update CrateDB used in tests to 4.0.2

mfussenegger:


chaudum:
I guess we need to update the Travis CI configuration so it uses JDK 12

mfussenegger:
> I guess we need to update the Travis CI configuration so it uses JDK 12

I think that's Ubuntu's openjdk-11 not really being openjdk 11 but an earlier version :(

I'm thinking about switching the repo to azure pipelines. What do you think?

chaudum:
> > I guess we need to update the Travis CI configuration so it uses JDK 12
> 
> I think that's Ubuntu's openjdk-11 not really being openjdk 11 but an earlier version :(
> 
> I'm thinking about switching the repo to azure pipelines. What do you think?

Sounds good to me!

src/crate/client/sqlalchemy/dialect.py
```diff
@@ -39,26 +39,38 @@
...
 }
 try:
     # SQLAlchemy >= 1.1
     from sqlalchemy.types import ARRAY
     TYPES_MAP["integer_array"] = ARRAY(sqltypes.Integer)
     TYPES_MAP["boolean_array"] = ARRAY(sqltypes.Boolean)
     TYPES_MAP["short_array"] = ARRAY(sqltypes.SmallInteger)
+    TYPES_MAP["smallint_array"] = ARRAY(sqltypes.SmallInteger)
     TYPES_MAP["timestamp_array"] = ARRAY(sqltypes.TIMESTAMP)
+    TYPES_MAP["timestamp with time zone_array"] = ARRAY(sqltypes.TIMESTAMP)
```
chaudum: That _array suffix looks strange when used with types that contain whitespaces.

mfussenegger: Yes, but that's what is returned by CrateDB

PR #325, opened 2019-07-15
Migrate CI to azure pipelines

mfussenegger:
Tends to be a bit faster and comes with more recent JDKs by default.


jodok:
last time i took a look at the pipelines i realized that they are running in an azure account that is separate from our main account. maybe this has changed or i looked the wrong way at it. but if not - any plans to move them (before we expand usage)?

chaudum:
The Azure pipeline jobs are missing the "post success" task of uploading the coverage report, as in https://github.com/crate/crate-python/pull/325/files#diff-354f30a63fb0907d4ad57269548329e3L46
Do we still want that?

mfussenegger:
@chaudum I'll follow up on that, possibly switching to codecov

PR #329, opened 2019-08-08
Add CODEOWNERS file

mechanomi:
this will auto-request a review from the @crate/docs team (but a review from the team is not mandatory)

### DIFF
--- a/.github/CODEOWNERS
+++ b/.github/CODEOWNERS
@@ -1,0 +1,2 @@
+*.rst @crate/docs
+docs/ @crate/docs


mechanomi:
@mkleen we have a few repos using this feature

I want to add this because I want the docs team to get notifications when documentation is updated

engineers often do not request a review from the docs team when making documentation changes. which means that errors can go unnoticed for quite some time.

requiring a docs team review on every PR is too much work. so this is sort of a middle-ground approach. it means we get a notification and can decide to weigh in if we want, but it's not required

this was the result of a conversation with @joemoe  about what to do to better enable oversight for our public-facing docs projects

mechanomi:
retest this please

PR #331, opened 2019-09-19
Default JAVA_HOME env in test layer to empty string

chaudum:

This commit changes the default value for `JAVA_HOME` in the
`CrateLayer` from `None` to `''` (empty string). This is necessary,
because environment variables are always interpreted as strings, and
therefore `None` would be interpreted as a string `'None'`, which is not
a valid path for `JAVA_HOME`.

## Checklist

### DIFF
--- a/CHANGES.txt
+++ b/CHANGES.txt
@@ -5,6 +5,9 @@ Changes for crate
 Unreleased
 ==========
 
+- Fixed a bug in the ``CrateLayer`` which caused ``CrateDB`` not to start up,
+  in case the ``JAVA_HOME`` environment variable was not set.
+
 2019/08/01 0.23.1
 =================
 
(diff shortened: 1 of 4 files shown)


src/crate/testing/test_layer.py
```diff
@@ -55,3 +57,28 @@ def test_wait_for_http(self):
         log = BytesIO(b'[i.c.p.h.CrateNettyHttpServerTransport] [crate] publish_address {}')
         addr = wait_for_http_url(log=log, timeout=1)
         self.assertEqual(None, addr)
+
+    @mock.patch.dict('os.environ')
+    def test_java_home_env_not_set(self):
+        with tempfile.TemporaryDirectory() as tmpdir:
+            layer = CrateLayer('java-home-test', tmpdir)
```
seut: I think the env must be unset here explicitly

CHANGES.txt
```diff
@@ -5,6 +5,9 @@ Changes for crate
 Unreleased
 ==========
 
+- Fixed a bug in the ``CrateLayer`` which caused ``CrateDB`` not to start up,
+  in case the ``JAVA_ENV`` environment variable was not set.
```
chaudum:
```suggestion
  in case the ``JAVA_HOME`` environment variable was not set.
```

PR #334, opened 2019-09-26
Register date_trunc function for crate dialect with correct return type

autophagy:

The `date_trunc` function in CrateDB returns a truncated timestamp based on what period you wish to truncate it to (minute, day, week, etc). This function only works on timestamp columns and returns a timestamp as its result.

However, when calling this function in SQLAlchemy, the returned type is not a `TIMESTAMP` but actually a `NullValue`, as SQLAlchemy is unaware of what the function should return. This autocasts the result set to a number, rather than a python `datetime`.

```
Base = declarative.declarative_base(bind=engine)

class Example(Base):
    __tablename__ = 't1'
    a = sa.Column("a", sa.DateTime, primary_key=True)

events = session.query(sa.func.date_trunc("minute", Example.a))
for r in events.all():
   print(type(r[0]))

>>> <class 'int'>
```

This change, therefore, registers the `date_trunc` function with the CrateDB dialect so that it returns a `TIMESTAMP`, rather than a generic null type:

```
events = session.query(sa.func.date_trunc("minute", Example.a))
for r in events.all():
   print(type(r[0]))

>>> <class 'datetime.datetime'>
```

This commit also adds a test for this case. This could be worth extending to other CrateDB functions that return specific types.

## Checklist

### DIFF
--- a/src/crate/client/sqlalchemy/tests/function_test.py
+++ b/src/crate/client/sqlalchemy/tests/function_test.py
@@ -1,0 +1,44 @@
+# -*- coding: utf-8; -*-
+#
+# Licensed to CRATE Technology GmbH ("Crate") under one or more contributor
+# license agreements.  See the NOTICE file distributed with this work for
+# additional information regarding copyright ownership.  Crate licenses
+# this file to you under the Apache License, Version 2.0 (the "License");
+# you may not use this file except in compliance with the License.  You may
+# obtain a copy of the License at
+#
+#   http://www.apache.org/licenses/LICENSE-2.0
+#
+# Unless required by applicable law or agreed to in writing, software
+# distributed under the License is distributed on an "AS IS" BASIS, WITHOUT
+# WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.  See the
+# License for the specific language governing permissions and limitations
+# under the License.
+#
+# However, if you have executed another commercial license agreement
+# with Crate these terms will supersede the license and you may use the
+# software solely pursuant to the terms of the relevant commercial agreement.
+
+from unittest import TestCase
+
+import sqlalchemy as sa
+from sqlalchemy.sql.sqltypes import TIMESTAMP
+from sqlalchemy.ext.declarative import declarative_base
+
+
+class FunctionTest(TestCase):
+    def setUp(self):
+        Base = declarative_base(bind=sa.create_engine("crate://"))
+
+        class Character(Base):
+            __tablename__ = "characters"
+            name = sa.Column(sa.String, primary_key=True)
+            timestamp = sa.Column(sa.DateTime)
+
+        self.Character = Character
+
+    def test_date_trunc_type_is_timestamp(self):
+        f = sa.func.date_trunc("minute", self.Character.timestamp)
+        self.assertEqual(len(f.base_columns), 1)
+        for col in f.base_columns:
+            self.assertIsInstance(col.type, TIMESTAMP)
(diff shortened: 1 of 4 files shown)


autophagy:
@chaudum Does this count as a bugfix, and so should be backported to 0.23? I'll also add an entry to the changes.

src/crate/client/sqlalchemy/tests/function_test.py
```diff
@@ -0,0 +1,55 @@
...
+
+import sqlalchemy as sa
+from sqlalchemy.sql.sqltypes import TIMESTAMP
+from sqlalchemy.ext.declarative import declarative_base
+from sqlalchemy.orm import Session
+
+fake_cursor = MagicMock(name="fake_cursor")
+
+
+class DialectTest(TestCase):
```
chaudum: This test doe not seem to be executed:

```
$ bin/test --list-tests | grep test_date_trunc_type_is_timestamp
```

src/crate/client/sqlalchemy/tests/function_test.py
```diff
@@ -0,0 +1,55 @@
...
+            __tablename__ = "characters"
+            name = sa.Column(sa.String, primary_key=True)
+            date = sa.Column(sa.Date)
+            timestamp = sa.Column(sa.DateTime)
+
+        fake_cursor.description = (
+            ("characters_name", None, None, None, None, None, None),
+            ("characters_date", None, None, None, None, None, None),
+        )
+        self.session = Session()
```
chaudum: `self.session` is not used anywhere else, so it can be omitted.

src/crate/client/sqlalchemy/tests/function_test.py
```diff
@@ -0,0 +1,55 @@
...
+        self.engine = sa.create_engine("crate://")
+        Base = declarative_base(bind=self.engine)
+
+        class Character(Base):
+            __tablename__ = "characters"
+            name = sa.Column(sa.String, primary_key=True)
+            date = sa.Column(sa.Date)
+            timestamp = sa.Column(sa.DateTime)
+
+        fake_cursor.description = (
```
chaudum: fake_cursor is unused.

src/crate/client/sqlalchemy/tests/function_test.py
```diff
@@ -0,0 +1,55 @@
...
+from sqlalchemy.sql.sqltypes import TIMESTAMP
+from sqlalchemy.ext.declarative import declarative_base
+from sqlalchemy.orm import Session
+
+fake_cursor = MagicMock(name="fake_cursor")
+
+
+class DialectTest(TestCase):
+    def setUp(self):
+        self.engine = sa.create_engine("crate://")
```
chaudum:
```suggestion
        engine = sa.create_engine("crate://")
```

src/crate/client/sqlalchemy/tests/function_test.py
autophagy: Good shout - I ran the test manually, but forgot to add it to the zope test suite. Will fix.

src/crate/client/sqlalchemy/tests/function_test.py
```diff
@@ -0,0 +1,45 @@
...
+
+
+class FunctionTest(TestCase):
+    def setUp(self):
+        Base = declarative_base(bind=sa.create_engine("crate://"))
+
+        class Character(Base):
+            __tablename__ = "characters"
+            name = sa.Column(sa.String, primary_key=True)
+            date = sa.Column(sa.Date)
```
chaudum: Just saw that `date` is also not relevant for the current test case.
`name` must be kept, because, a SQLAlchemy model requires a primary key.

Issue #339, opened 2020-02-03
Issue in Connection timeout when CrateDB is not available


daminichopra:
I am using CrateDb for my application but facing an issue while receiving a notification, if the DB isn't available then it returns **500 Internal Error** quickly.
 CrateDb  has no retry interval between total retries. So, i suggest that CrateDb should wait for certain time between retries and do not return result immediately.

As per my analysis, the  parameter [**"backoff_factor"**](https://urllib3.readthedocs.io/en/latest/reference/urllib3.util.html) is defined in crateDB, which is used to define the time-interval between consecutive retries. The default value of **backoff_factor** can be assigned as **"0.0"**  and it can be changed as per requirement.
I am using CrateDb version : 3.1.2

Also, Please find attached logs for the same.
![image](https://user-images.githubusercontent.com/35693908/73639842-2e59df00-4693-11ea-941f-d6b42d13932c.png)
![image](https://user-images.githubusercontent.com/35693908/73639852-34e85680-4693-11ea-831e-f7f47c5782bc.png)

In reference to following [Issue](https://github.com/smartsdk/ngsi-timeseries-api/issues/242).

daminichopra:
Hi @autophagy , As i wanted the solution for this issue urgently and i have already fixed the code so can i raise the PR for the same that it can be closed as soon as possible. Please confirm. Thank you

mariacrate:
Hi @daminichopra.
Thanks for raising this subject. We're very pleased CrateDB supports your application. We would be glad to assist you further with the issue you're struggling. Please feel free to share your contact details and I will be back with you shortly. Also, you can reach me out directly at maria@crate.io
I look forward to hearing from you soon!
Maria

daminichopra:
Hi @autophagy , I have fixed the code in cratedb. Now , i want to test code locally for which i am using pytest but test cases are not running.
Can you please guide me for the steps on how to run test cases in crateDB.
Thank you :)

mfussenegger:
> 
> 
> Hi @autophagy , I have fixed the code in cratedb. Now , i want to test code locally for which i am using pytest but test cases are not running.
> Can you please guide me for the steps on how to run test cases in crateDB.
> Thank you :)

This project currently uses the zope testrunner, see [Running Tests](https://github.com/crate/crate-python/blob/master/DEVELOP.rst#running-tests) in our developer docs.

autophagy:
Fixed with #344

PR #340, opened 2020-02-04
Update code for retry interval between retries

daminichopra:
Updated code in http.py and introduced **backoff_factor=0.0** as this parameter can do retry interval between consecutive retries. So, that when user want to change value of **backoff_factor=0.0** to 0.0,0.1,0.2 ... for retry interval they can edit value of backoff_factor as per requirement.

## Checklist

 - [ OK ] [CLA](https://crate.io/community/contribute/cla/) is signed

### DIFF
--- a/src/crate/client/http.py
+++ b/src/crate/client/http.py
@@ -130,7 +130,7 @@ class Server(object):
         headers['Content-Type'] = 'application/json'
         kwargs['assert_same_host'] = False
         kwargs['redirect'] = False
-        kwargs['retries'] = Retry(read=0)
+        kwargs['retries'] = Retry(read=0,backoff_factor=0.0)
         return self.pool.urlopen(
             method,
             path,


autophagy:
@daminichopra Hi Damini, thank you for submitting a pull request. :slightly_smiling_face: 

If I'm correct, this PR is related to #339, which was about the lack of a time backoff when retrying an unavailable CrateDB server? According to the documentation for `urllib3`, "by default, backoff is disabled (set to 0)." https://urllib3.readthedocs.io/en/latest/reference/urllib3.util.html#module-urllib3.util.retry

Since this is already 0 by default, I'm not sure setting it as 0 would have the desired effect. What is the effect you wanted this change to have? Would you be able to explain the intention behind this change more in the commit message? Thank you!

daminichopra:
Hi @autophagy , I have updated the summary for the PR. Please check.Thank you :)

> @daminichopra Hi Damini, thank you for submitting a pull request. 
> 
> If I'm correct, this PR is related to #339, which was about the lack of a time backoff when retrying an unavailable CrateDB server? According to the documentation for `urllib3`, "by default, backoff is disabled (set to 0)." https://urllib3.readthedocs.io/en/latest/reference/urllib3.util.html#module-urllib3.util.retry
> 
> Since this is already 0 by default, I'm not sure setting it as 0 would have the desired effect. What is the effect you wanted this change to have? Would you be able to explain the intention behind this change more in the commit message? Thank you!

autophagy:
@daminichopra Thank you for the response. I'm still a little confused :sweat_smile: how would another user go about changing this value? Directly in the code?

daminichopra:
Hi @autophagy , yes you are right. To change value of "backoff_factor", i think we need to make it configurable in crateDb code by initializing new parameter "backoff_factor" . If you have any other opinion then Please suggest. Thank you :)

daminichopra:
In refrence to PR#[342](https://github.com/crate/crate-python/pull/342)

PR #373, opened 2020-09-07
Make connection pool size configurable

amotl:
Hi there,

we have been observing several warnings like these on our live environments:
```
WARNING  [urllib3.connectionpool          ] Connection pool is full, discarding connection: testdrive.cratedb.example.net
```

## Reproduction
In order to reproduce these conveniently, you might want to use [`crate-parallel.py`](https://gist.github.com/amotl/575aae129beb17b120b848e7632365dc). It will invoke `SELECT 1;` 100 times in parallel on 10 worker threads and will immediately trigger these warnings. When invoked with the `--pool` parameter - essentially emulating this PR - everything goes along without warnings.

## Background
We found out that the [default pool size of `urllib3` is just "1"](https://github.com/urllib3/urllib3/blob/1.25.10/src/urllib3/connectionpool.py#L178), while with the [`requests` module, it is "10" already](https://github.com/psf/requests/blob/v2.24.0/requests/adapters.py#L50).

## Mitigation
- This update makes the connection pool size configurable, which is important in multi-threaded situations.
- `client.connect()` will accept a new parameter `pool_size` and will propagate it to urllib3 as `maxsize` parameter.
- ~While "urllib3" used a default of 1, the new default is now 10, following the default of the "requests" module.~

With kind regards,
Andreas.

### DIFF
--- a/src/crate/client/connection.py
+++ b/src/crate/client/connection.py
@@ -28,10 +28,20 @@ from distutils.version import StrictVersion
 
 class Connection(object):
 
-    def __init__(self, servers=None, timeout=None, backoff_factor=0, client=None,
-                 verify_ssl_cert=False, ca_cert=None, error_trace=False,
-                 cert_file=None, key_file=None, username=None, password=None,
-                 schema=None):
+    def __init__(self,
+                 servers=None,
+                 timeout=None,
+                 backoff_factor=0,
+                 client=None,
+                 verify_ssl_cert=False,
+                 ca_cert=None,
+                 error_trace=False,
+                 cert_file=None,
+                 key_file=None,
+                 username=None,
+                 password=None,
+                 schema=None,
+                 pool_size=None):
         if client:
             self.client = client
         else:
@@ -45,7 +55,8 @@ class Connection(object):
                                  key_file=key_file,
                                  username=username,
                                  password=password,
-                                 schema=schema)
+                                 schema=schema,
+                                 pool_size=pool_size)
         self.lowest_server_version = self._lowest_server_version()
         self._closed = False
 
@@ -113,7 +124,8 @@ def connect(servers=None,
             key_file=None,
             username=None,
             password=None,
-            schema=None):
+            schema=None,
+            pool_size=None):
     """ Create a :class:Connection object
 
     :param servers:
@@ -122,6 +134,9 @@ def connect(servers=None,
     :param timeout:
         (optional)
         define the retry timeout for unreachable servers in seconds
+    :param backoff_factor:
+        (optional)
+        define the retry interval for unreachable servers in seconds
     :param client:
         (optional - for testing)
         client used to communicate with crate.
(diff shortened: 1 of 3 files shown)


src/crate/client/connection.py
```diff
@@ -28,16 +28,27 @@
...
 class Connection(object):
 
-    def __init__(self, servers=None, timeout=None, backoff_factor=0, client=None,
-                 verify_ssl_cert=False, ca_cert=None, error_trace=False,
-                 cert_file=None, key_file=None, username=None, password=None,
+    def __init__(self,
+                 servers=None,
+                 timeout=None,
+                 backoff_factor=0,
+                 poolsize=10,
```
mfussenegger, line 35: Could you move `poolsize` to the last position to not break BWC with calls that use positional arguments?

src/crate/client/connection.py
```diff
@@ -105,6 +116,7 @@ def __exit__(self, *excs):
 def connect(servers=None,
             timeout=None,
             backoff_factor=0,
+            poolsize=10,
```
mfussenegger, line 119: Same here, please move to the last position.

src/crate/client/http.py
```diff
@@ -276,6 +276,7 @@ def __init__(self,
                  servers=None,
                  timeout=None,
                  backoff_factor=0,
+                 poolsize=10,
```
mfussenegger, line 279: Same here

mfussenegger (commented):
Left some comments about keeping BWC - overall the change makes sense to me :+1: 


Not sure about the default value. Personally I'd stick with the urllib3 default instead of going with what requests chose. Mostly because I dislike changing upstream defaults without very good reason. Not sure if the  "we think threaded uese-cases where a client makes connections against multiple servers is more common" counts as such. But no strong opinion.

src/crate/client/connection.py
amotl: Sure. Done.

src/crate/client/connection.py
amotl: Thanks again!

src/crate/client/http.py
amotl: Dito.

chaudum:
I tend to agree with @mfussenegger that we should keep the upstream defaults as they are and change them only if the `pool_size` parameter is set.

amotl:
Dear @mfussenegger and @chaudum,

I've updated this PR according to your suggestions.

With kind regards,
Andreas.

mfussenegger (approved):

src/crate/client/connection.py
```diff
@@ -122,6 +134,13 @@ def connect(servers=None,
     :param timeout:
         (optional)
         define the retry timeout for unreachable servers in seconds
+    :param backoff_factor:
+        (optional)
+        define the retry interval for unreachable servers in seconds
+    :param pool_size:
+        (optional)
+        Number of connections to save that can be reused.
+        More than 1 is useful in multithreaded situations.
```
chaudum, line 143: I would adjust the order of the parameter documentation to the order of the arguments (move at the end).

chaudum (approved):
Otherwise lgtm

src/crate/client/connection.py
amotl: Thanks. I've adjusted the parameter order within the documentation as suggested.

amotl:
That was quick. Thanks for merging!

amotl:
Hi again,

when using `crate-python` within a multi-threaded environment, it is important to configure the `pool_size` option appropriately. The gist at [1] outlines this in both vanilla DBAPI- and SQLAlchemy-scenarios.

We might want to add this to documentation somehow.

With kind regards,
Andreas.

cc @chaudum, @mfussenegger, @autophagy, @mechanomi 

[1] https://gist.github.com/amotl/575aae129beb17b120b848e7632365dc

mechanomi:
@amotl that gist is awesome. do you think it would make sense to add it to the crate-python docs as a how-to guide? I would be more than happy to edit this with you so it can be published

amotl:
Hi Naomi,

> Do you think it would make sense to add [this gist] to the crate-python docs as a how-to guide?

I do. I believe it is very important to give users an insight in how to properly use crate-python from multi-threaded environments. I would appreciate if this wouldn't get lost within my gist and I would appreciate even more if you can lend a hand on this.

Cheers,
Andreas.

PR #374, opened 2020-09-09
Enable TCP keep alive on socket level

chaudum:

`urllib3` allows to set socket options for individual connections.

Reference: https://urllib3.readthedocs.io/en/latest/reference/urllib3.connection.html#urllib3.connection.HTTPConnection

This should help mitigating problems with environments where idle connections are closed without resetting. This happens e.g. with Azure load balancers.

Reference: https://joonas.fi/2017/01/23/microsoft-azures-networking-is-fundamentally-broken/

## Checklist

 - [x] Add documentation about socket options

### DIFF
--- a/docs/connect.rst
+++ b/docs/connect.rst
@@ -167,19 +167,50 @@ the optional ``error_trace`` argument to ``True``, like so::
 
     >>> connection = client.connect(..., error_trace=True)
 
-.. _authentication:
-
 Backoff Factor
-..............
+--------------
 
 When attempting to make a request, the connection can be configured so that
 retries are made in increasing time intervals. This can be configured like so::
 
     >>> connection = client.connect(..., backoff_factor=0.1)
 
-If ``backoff_factor``is set to 0.1, then the delay between retries will be 0.0,
+If ``backoff_factor`` is set to 0.1, then the delay between retries will be 0.0,
 0.1, 0.2, 0.4 etc. The maximum backoff factor cannot exceed 120 seconds and by
 default its value is 0.
+
+Socket Options
+--------------
+
+Creating connections uses `urllib3 default socket options`_ but additionally
+enables TCP keepalive by setting ``socket.SO_KEEPALIVE`` to ``1``.
+
+Keepalive can be disabled using the ``socket_keepalive`` argument, like so::
+
+    >>> connection = client.connect(..., socket_keepalive=False)
+
+If keepalive is enabled (default), there are three additional, optional socket
+options that can be configured via connection arguments.
+
+:``socket_tcp_keepidle``:
+
+    Set the ``TCP_KEEPIDLE`` socket option, which overrides
+    ``net.ipv4.tcp_keepalive_time`` kernel setting if ``socket_keepalive`` is
+    ``True``.
+
+:``socket_tcp_keepintvl``:
+
+    Set the ``TCP_KEEPINTVL`` socket option, which overrides
+    ``net.ipv4.tcp_keepalive_intvl`` kernel setting if ``socket_keepalive`` is
+    ``True``.
+
+:``socket_tcp_keepcnt``:
+
+    Set the ``TCP_KEEPCNT`` socket option, which overrides
+    ``net.ipv4.tcp_keepalive_probes`` kernel setting if ``socket_keepalive`` is
+    ``True``.
+
+.. _authentication:
 
 Authentication
 ==============
@@ -247,3 +278,4 @@ Once you're connected, you can :ref:`query CrateDB <query>`.
 .. _socket timeout: https://docs.python.org/2/library/socket.html#socket.getdefaulttimeout
 .. _SQLAlchemy: http://www.sqlalchemy.org/
 .. _tracebacks: https://docs.python.org/3/library/traceback.html
+.. _urllib3 default socket options: https://urllib3.readthedocs.io/en/latest/reference/urllib3.connection.html#urllib3.connection.HTTPConnection
--- a/src/crate/client/http.py
+++ b/src/crate/client/http.py
@@ -20,28 +20,37 @@
 # software solely pursuant to the terms of the relevant commercial agreement.
 
 
+import calendar
 import heapq
+import io
 import json
 import logging
 import os
-import io
+import re
+import socket
 import ssl
-import urllib3
-import urllib3.exceptions
-from urllib3.util.retry import Retry
+import threading
+from urllib.parse import urlparse
 from base64 import b64encode
 from time import time
 from datetime import datetime, date
 from decimal import Decimal
-import calendar
-import threading
-import re
-from urllib.parse import urlparse
+from urllib3 import connection_from_url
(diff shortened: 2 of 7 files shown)


mfussenegger (commented):
Should we also change the value for `socket.TCP_KEEPIDLE` ?

(Based on some comments elsewhere that the default interval is larger than the ~4m that azure has)

amotl:
Dear Christian,

thanks for working on this. As suggested on [Azure closing idle network connections](https://serverfault.com/a/851251) or [TCP settings for Azure VMs](https://social.msdn.microsoft.com/Forums/en-US/0462f436-2493-4199-ba1b-e6d05fdaebd4/azure-vm-tcp-idle-timeout?forum=WAVirtualMachinesforWindows#1cca8e0d-43a8-4c15-a946-12de725f2934), the recommended Kernel TCP settings for mitigating the named issue on Azure are
```
net.ipv4.tcp_keepalive_time = 120
net.ipv4.tcp_keepalive_intvl = 30
net.ipv4.tcp_keepalive_probes = 8
```

According to [TCP Keepalive HOWTO » The setsockopt function call](https://tldp.org/HOWTO/html_single/TCP-Keepalive-HOWTO/#setsockopt),

> There are also three other socket options you can set for keepalive when you write your application. They all use the `SOL_TCP` level instead of `SOL_SOCKET`, and they override system-wide variables only for the current socket. If you read without writing first, the current system-wide parameters will be returned.
> 
> - `TCP_KEEPIDLE`: overrides `tcp_keepalive_time`
> - `TCP_KEEPINTVL`: overrides `tcp_keepalive_intvl`
> - `TCP_KEEPCNT`: overrides `tcp_keepalive_probes`

@mfussenegger asked:
> Should we also change the value for `socket.TCP_KEEPIDLE`?

When we believe the eventual issue with Azure networking is something worthy to consider and we prefer to do it on the application level, we might well want to add configuration capabilities for these three socket option values, too.

With kind regards,
Andreas.

amotl:
The article [When TCP sockets refuse to die](https://slacker.ro/2019/09/20/when-tcp-sockets-refuse-to-die/) also talks about the fairly modern `TCP_USER_TIMEOUT` socket option. They say it is equally as important. I didn't have a clue about this option before.

chaudum:
> When we believe the eventual issue with Azure networking is something worthy to consider and we prefer to do it on the application level, we might well want to add configuration capabilities for these three socket option values, too.

I think it makes sense to make these options configurable. I implemented it in a way that keepalive (`socket.SO_KEEPALIVE`) is enabled by default. If enabled, the kernel settings can be overwritten by individual options. Setting the socket options is still handled by `urllib3`.

src/crate/client/http.py
```diff
@@ -258,6 +283,28 @@ def _create_sql_payload(stmt, args, bulk_args):
...
+                     tcp_keepintvl=None,
+                     tcp_keepcnt=None):
+    """
+    Return an optional list of socket options for urllib3's HTTPConnection
+    constructor.
+    """
+    if not keepalive:
+        return None
+
+    opts = [(socket.SOL_SOCKET, socket.SO_KEEPALIVE, 1)]
```
amotl, line 297: Just a small nit to improve inline documentation about what we are essentially doing here.

```suggestion
    # Let's always use TCP keepalive.
    opts = [(socket.SOL_SOCKET, socket.SO_KEEPALIVE, 1)]

    # Optionally adjust TCP keepalive parameters.
```

amotl (approved):
That was quick. Thanks a bunch for your efforts!

chaudum:
> P.S.: Apparently, others also had issues with this (on Azure), see https://jira.mongodb.org/browse/DRIVERS-383. It is a specification all drivers should adhere to by applying and configuring TCP keepalive by default.

I like the idea that we do that as well. All http clients should use the same tcp socket options.

src/crate/client/http.py
mfussenegger: Not so sure if this improves things. It repeats what the code already tells a reader, but doesn't give a rationale. ("Document *why*, not *what*")

Something like this would imho make more sense.

```suggestion
    # Prevents issues where load balancers or ISPs terminate in-activate connections
    # See https://joonas.fi/2017/01/23/microsoft-azures-networking-is-fundamentally-broken/ 
    opts = [(socket.SOL_SOCKET, socket.SO_KEEPALIVE, 1)]
```

src/crate/client/http.py
```diff
@@ -258,6 +283,30 @@ def _create_sql_payload(stmt, args, bulk_args):
...
+    Return an optional list of socket options for urllib3's HTTPConnection
+    constructor.
+    """
+    if not keepalive:
+        return None
+
+    # Let's always use TCP keepalive.
+    opts = [(socket.SOL_SOCKET, socket.SO_KEEPALIVE, 1)]
+
+    if hasattr(socket, 'TCP_KEEPIDLE') and tcp_keepidle is not None:
```
mfussenegger, line 300:
```suggestion
    # hasattr check because some of the options depend on system capabilities
    # see https://docs.python.org/3/library/socket.html?highlight=socket#socket.SOMAXCONN
    if hasattr(socket, 'TCP_KEEPIDLE') and tcp_keepidle is not None:
```

mfussenegger (approved):
left some more minor comments.

Overall I wonder if we should consider adding some kind of server_factory so that users get full control over how the pool is initialized if they want to.

Somehow we end up exposing more and more options for urllib3 to the users.

chaudum:
> left some more minor comments.

Will integrate the suggestion.

> Overall I wonder if we should consider adding some kind of server_factory so that users get full control over how the pool is initialized if they want to.
> 
> Somehow we end up exposing more and more options for urllib3 to the users.

I agree, it's getting a bit messy. I did not want to do too much refactoring in this PR though. I will have a look how the situation can be improved.

chaudum:
@amotl @mfussenegger @crate/tech-writing Can you have a look at the additions to the documentation?

amotl:
> Can you have a look at the additions to the documentation?

lgtm. Thanks!

mechanomi (approved):

docs/connect.rst
```diff
@@ -167,20 +167,51 @@ the optional ``error_trace`` argument to ``True``, like so::
...
-If ``backoff_factor``is set to 0.1, then the delay between retries will be 0.0,
+If ``backoff_factor`` is set to 0.1, then the delay between retries will be 0.0,
 0.1, 0.2, 0.4 etc. The maximum backoff factor cannot exceed 120 seconds and by
 default its value is 0.
 
+Socket Options
+--------------
+
+Creating connections uses `urllib3 default socket options`_ but additionally
+add enable TCP keepalive by setting ``socket.SO_KEEPALIVE`` to ``1``.
```
mfussenegger, line 186:
```suggestion
Creating connections uses `urllib3 default socket options`_ but additionally
enables TCP keepalive by setting ``socket.SO_KEEPALIVE`` to ``1``.
```

mfussenegger (approved):

Issue #378, opened 2020-09-28
Deploy to Pypi via Github Action


autophagy:
At the moment, deploying a new crate python release to pypi is a manual step, where the developer has to `setup.py sdist bdist_wheel` and then upload the artifacts using `twine`.

It would be cool if we could use a github action to do this, similar to crash: https://github.com/crate/crash/blob/master/.github/workflows/main.yml#L37

PR #379, opened 2020-09-28
Add github action to release project to pypi on tag

autophagy:

This replaces the manual release of the Crate python driver to PyPi to one done by Github actions, triggered on tag pushes, similar to our Crash project. The pypi API key has already been added to the secrets for this project.

Addresses #378

### DIFF
--- a/.github/workflows/release.yml
+++ b/.github/workflows/release.yml
@@ -1,0 +1,29 @@
+---
+name: release
+
+on: push
+
+jobs:
+  pypi:
+    name: Build & publish package to pypi
+    runs-on: ubuntu-latest
+    if: startsWith(github.event.ref, 'refs/tags')
+    steps:
+      - uses: actions/checkout@master
+
+      - name: Set up python 3.8
+        uses: actions/setup-python@v1
+        with:
+          python-version: 3.8
+
+      - name: Build package
+        run: |
+          python -m pip install twine wheel
+          python setup.py sdist bdist_wheel
+          twine check dist/*.tar.gz
+
+      - name: Publish package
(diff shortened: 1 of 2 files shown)


autophagy:
Since this duplicates tests (as can be seen on this PR :grimacing:), maybe it would make sense to remove the old testing one and just have this one take its place? The GH action in crash does the same thing.

chaudum:
> Since this duplicates tests (as can be seen on this PR grimacing), maybe it would make sense to remove the old testing one and just have this one take its place? The GH action in crash does the same thing.

However, the old one also runs the tests periodically.

mfussenegger:
> Since this duplicates tests (as can be seen on this PR ), maybe it would make sense to remove the old testing one and just have this one take its place? The GH action in crash does the same thing.

Or we could run the new thing only on push against master instead of push on any branch, and skip the tests (they must have passed if something got merged)

autophagy:
Good points! I'll remove the testing and just run against master.

chaudum:
I am wondering whether the action is run, since it is only run on push to master branch. However, pushing a tag `git push --tags` does not push anything to master.

I double checked with how it's done for `croud` and there the action runs on all push events and filters on the event.ref: https://github.com/crate/croud/blob/master/.github/workflows/release.yaml#L7

mechanomi (commented):
lgtm

autophagy:
@chaudum Thank you for your help on this :slightly_smiling_face: I've pushed a fixup so the action is triggered by any push, like ``croud``.

chaudum (approved):

PR #380, opened 2020-10-13
Add python 3.9 to CI & list as supported

mfussenegger:

PR #383, opened 2020-10-22
Attempt to fix CI with Python 3.9

amotl:
This tries to fix `AttributeError: '_MainThread' object has no attribute 'isAlive'` CI was croaking on after adding support for Python 3.9, see also https://github.com/crate/crate-python/pull/380#issuecomment-714452850.


amotl:
This has been integrated into #380 already. Thanks!

PR #384, opened 2020-10-28
Fix failing https.txt test

autophagy:
This error occurs in some python version tests but not in others - I believe the actual exception message differs across different library versions, but the outcome is ultimately the same.

### DIFF
--- a/src/crate/client/doctests/https.txt
+++ b/src/crate/client/doctests/https.txt
@@ -76,7 +76,7 @@ certificate and key file.
 Below an example, in this case it fails because the supplied certificate is
 invalid::
 
-    >>> client = HttpClient([crate_host], cert_file=invalid_ca_cert, key_file=invalid_ca_cert)
+    >>> client = HttpClient([crate_host], cert_file=invalid_ca_cert, key_file=invalid_ca_cert, verify_ssl_cert=True)
     >>> client.server_infos(crate_host)
     Traceback (most recent call last):
     ...


amotl:
After just integrating #387, I've (accidentally/short-sightedly) pressed the "Update branch" button here. I don't know if that was the right decision as it created a merge commit here.

Sorry so much if that will create any hassle for you. We might want to rebase onto master and force push here again if that might resolve something.

src/crate/client/doctests/https.txt
```diff
@@ -80,5 +80,5 @@ invalid::
     >>> client.server_infos(crate_host)
     Traceback (most recent call last):
     ...
-    crate.client.exceptions.ConnectionError: Server not available, exception: ...[SSL: ...
+    crate.client.exceptions.ConnectionError: Server not available, exception:...
```
chaudum, line 83: Is there anything else in the traceback that would indicate that the certificate is invalid? Asserting only the `ConnectionError` seems a bit generic.

amotl:
While I haven't been able to spot this error when running the test suite on my machine, I received respective hiccups on CI and reported about them at https://github.com/crate/crate-python/pull/386#issuecomment-718114535.

@autophagy: Can you look at the comment by @chaudum and then let us aim at getting this merged?

src/crate/client/doctests/https.txt
amotl: On this matter, I just received this error with traceback (from [1]):
```
File "D:\a\crate-python\crate-python\src\crate\client\doctests\https.txt", line 80, in https.txt
Failed example:
    client.server_infos(crate_host)
Expected:
    ...
    crate.client.exceptions.ConnectionError: Server not available, exception: ...[SSL: ...
Got:
      File "C:\hostedtoolcache\windows\Python\3.7.9\x64\lib\doctest.py", line 1337, in __run
        compileflags, 1), test.globs)
      File "<doctest https.txt[11]>", line 1, in <module>
        client.server_infos(crate_host)
      File "D:\a\crate-python\crate-python\src\crate\client\http.py", line 402, in server_infos
        response = self._request('GET', '/', server=server)
      File "D:\a\crate-python\crate-python\src\crate\client\http.py", line 502, in _request
        "Server not available, exception: %s" % ex_message
    crate.client.exceptions.ConnectionError: Server not available, exception: HTTPSConnectionPool(host='localhost', port=65534): Max retries exceeded with url: / (Caused by ProtocolError('Connection aborted.', ConnectionResetError(10054, 'An existing connection was forcibly closed by the remote host', None, 10054, None)))
```

[1] https://github.com/crate/crate-python/pull/388/checks?check_run_id=1348073153#step:5:156

amotl: > unless this error got masked by any other phenomenon

Forget about my last comment. The reason for the `ConnectionResetError` is coming from the fact that CrateDB has not spun up completely, see also https://github.com/crate/crate-python/pull/388#issuecomment-721191227.

autophagy: The problem is, this failing https test doesnt actually set up a cratedb cluster to do it. It just creates a generic https server and attempts to query it with the client, see: https://github.com/crate/crate-python/blob/master/src/crate/client/tests.py#L210

autophagy:
@chaudum @amotl Sorry for not looking into this a little deeper - i've pushed a commit that should fix the test without modifying the expected error message.

src/crate/client/doctests/https.txt
amotl: I see. Thanks!

amotl:
> I've pushed a commit that should fix the test without modifying the expected error message.

Ha. Was it really just about adding `verify_ssl_cert=True`?

amotl:
We can merge this anyway and see if the issue will come back again. Ack.

amotl (approved):

PR #391, opened 2020-11-16
Add support for SQLAlchemy 1.4

amotl:
SQLAlchemy 1.4.0 is about to be released. This is just a testdrive to check whether things will break or not.

### DIFF
--- a/.github/workflows/main.yml
+++ b/.github/workflows/main.yml
@@ -21,7 +21,7 @@ jobs:
       matrix:
         crate-version: [nightly]
         os: [ubuntu-latest]
-        sqla-version: ['1.1.18', '1.2.19', '1.3.20']
+        sqla-version: ['1.1.18', '1.2.19', '1.3.20', '1.4.0b1']
         python-version: [3.5, 3.6, 3.7, 3.8, 3.9]
 
     steps:
--- a/setup.py
+++ b/setup.py
@@ -66,7 +66,7 @@ setup(
     extras_require=dict(
         test=['zope.testing',
               'zc.customdoctests>=1.0.1'],
-        sqlalchemy=['sqlalchemy>=1.0,<1.4', 'geojson>=2.5.0']
+        sqlalchemy=['sqlalchemy>=1.0,<1.5', 'geojson>=2.5.0']
     ),
     python_requires='>=3.4',
     install_requires=requirements,
--- a/tox.ini
+++ b/tox.ini
@@ -1,5 +1,5 @@
 [tox]
-envlist = py{py,34,35,36,37,38}-sa_{1_0,1_1,1_2,1_3}
+envlist = py{py,34,35,36,37,38}-sa_{1_0,1_1,1_2,1_3,1_4}
 
 [testenv]
 usedevelop = True
@@ -12,6 +12,7 @@ deps =
     sa_1_1: sqlalchemy>=1.1,<1.2
     sa_1_2: sqlalchemy>=1.2,<1.3
     sa_1_3: sqlalchemy>=1.3,<1.4
+    sa_1_4: sqlalchemy>=1.4,<1.5
     mock
     urllib3
 commands =


amotl:
SQLAlchemy 1.4.0 will drop support for Python 3.5, see https://github.com/sqlalchemy/sqlalchemy/commit/d0b5ce2ad4cb6149dbe3a378665c9c093da3f4f7.

amotl:
### Observations
Both `itest.txt` and `sqlalchemy.txt` fail with:
```
      File "/home/runner/work/crate-python/crate-python/eggs/SQLAlchemy-1.4.0b1-py3.8-linux-x86_64.egg/sqlalchemy/sql/compiler.py", line 508, in process
        return obj._compiler_dispatch(self, **kwargs)
      File "<string>", line 9, in _compiler_dispatch
      File "/home/runner/work/crate-python/crate-python/src/crate/client/sqlalchemy/compiler.py", line 201, in visit_insert
        crud_params = crud._get_crud_params(self, insert_stmt, **kw)
    TypeError: _get_crud_params() missing 1 required positional argument: 'compile_state'
```

-- https://github.com/crate/crate-python/pull/391/checks?check_run_id=1406879590

### Root cause
The signature of `_get_crud_params()` has been updated through https://github.com/sqlalchemy/sqlalchemy/commit/851fb8f5a661c66ee76308181118369c8c4df9e0, which decouples compiler state from DML objects.

amotl:
Dear @chaudum and @m-kharbat,

I've tried to apply some minor updates from SQLAlchemy 1.4 to our `visit_insert()` and `visit_update()` methods.

https://github.com/crate/crate-python/blob/7e12a49d46da36d5f502db38bfc780793b50b786/src/crate/client/sqlalchemy/compiler.py#L187-L193

https://github.com/crate/crate-python/blob/7e12a49d46da36d5f502db38bfc780793b50b786/src/crate/client/sqlalchemy/compiler.py#L285-L289

However, after I didn't have quick success when making a few adjustments and after reading the comments above, I wanted to check back with you about the further plan. Regarding the _conform with earlier versions of CreateDB_ comment: Maybe we don't need any amendments coming from these functions anymore and drop support for CrateDB 0.x releases?

Apart from that, I also found that some bits within these methods are apparently related to still support SQLAlchemy 1.0 and 1.1. Maybe we also want to revise this and drop respective support?

With kind regards,
Andreas.

amotl (commented):
39d0101, after some other minor adjustments, magically made the test suite succeed, by adjusting the parameter mangling to/from [`crate_before_execute`](https://github.com/crate/crate-python/blob/0.26.0/src/crate/client/sqlalchemy/compiler.py#L73-L78).

Other than the inline comments below, some other items could/should be addressed before finalizing this patch. Please let me know about your suggestions and if you can spot other shortcomings.

1. Special attention should be given to the new SQL Compilation Caching feature [1,2]. The `CrateDialect` should `supports_statement_cache = True`, see [3]. Also, there is a new `cache_ok` attribute on custom types [4], which should probably be applied to all CrateDB SQLAlchemy types listed in `crate.client.sqlalchemy.types`. [5] might also be related here.

   Currently, we observe all those warnings referenced at SA's [identify what constructs are blocking caching from being enabled](https://docs.sqlalchemy.org/en/14/faq/performance.html#step-two-identify-what-constructs-are-blocking-caching-from-being-enabled) documentation:
   > SAWarning: Dialect crate:crate-python will not make use of SQL compilation caching as it does not set the `supports_statement_cache` attribute to `True`.  This can have significant performance implications including some performance degradations in comparison to prior SQLAlchemy versions.  Dialect maintainers should seek to set this attribute to True after appropriate development and testing for SQLAlchemy 1.4 caching support.   Alternatively, this attribute may be set to `False` which will disable this warning. (Background on this error at: https://sqlalche.me/e/14/cprf)

   > SAWarning: UserDefinedType `{_Craty(),_ObjectArray(),Geopoint(),Geoshape()}` will not produce a cache key because the ``cache_ok`` attribute is not set to `True`.  This can have significant performance implications including some performance degradations in comparison to prior SQLAlchemy versions.  Set this attribute to `True` if this type object's state is safe to use in a cache key, or `False` to disable this warning. (Background on this error at: https://sqlalche.me/e/14/cprf)

   > SAWarning: `{Class Any,Class Match}` will not make use of SQL compilation caching as it does not set the `inherit_cache` attribute to `True`.  This can have significant performance implications including some performance degradations in comparison to prior SQLAlchemy versions.  Set this attribute to `True` if this object can make use of the cache key generated by the superclass.  Alternatively, this attribute may be set to `False` which will disable this warning. (Background on this error at: https://sqlalche.me/e/14/cprf)

2. Go through the list of significant changes [6] thoroughly and summarize/compile relevant details regarding CrateDB into the changelog file.

3. Test Asynchronous I/O with `asyncpg`, see [7].
   ```
   engine = create_async_engine(
     "postgresql+asyncpg://scott:tiger@localhost/test"
   )
   ```

[1] https://docs.sqlalchemy.org/en/14/core/connections.html#sql-compilation-caching
[2] https://docs.sqlalchemy.org/en/14/faq/performance.html#faq-new-caching
[3] https://docs.sqlalchemy.org/en/14/core/connections.html#caching-for-third-party-dialects
[4] https://docs.sqlalchemy.org/en/14/core/custom_types.html
[5] https://docs.sqlalchemy.org/en/14/errors.html#caching-caveats
[6] https://docs.sqlalchemy.org/en/14/changelog/migration_14.html
[7] https://docs.sqlalchemy.org/en/14/orm/extensions/asyncio.html

CHANGES.txt
```diff
@@ -16,6 +16,42 @@ Unreleased
...
+to compensate for apparent additional strictness of SQLAlchemy on some details.
+
+Where it was ok to use a textual column expression in plain text beforehand,
+a `SQLAlchemy literal_column`_ type should be used now. Otherwise, for example
+when accessing `CrateDB system columns`_ like ``_score``, the engine might
+complain like::
+
+    sqlalchemy.exc.ArgumentError: Textual column expression '_score' should be
+    explicitly declared with text('_score'), or use column('_score') for more
+    specificity
```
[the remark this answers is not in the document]
amotl: I've tried to improve it with 09b3173bd. Further suggestions are welcome.

src/crate/client/sqlalchemy/dialect.py
```diff
@@ -242,6 +261,15 @@ def get_table_names(self, connection, schema=None, **kw):
...
         return [row[0] for row in cursor.fetchall()]
 
+    @reflection.cache
+    def get_view_names(self, connection, schema=None, **kw):
+        cursor = connection.execute(
+            "SELECT table_name FROM information_schema.views "
+            "ORDER BY table_name ASC, {0} ASC".format(self.schema_column),
+            [schema or self.default_schema_name]
+        )
+        return [row[0] for row in cursor.fetchall()]
```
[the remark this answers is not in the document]
amotl: A test case has been added with c8656eec.

src/crate/client/sqlalchemy/monkey.py
```diff
@@ -0,0 +1,44 @@
...
+    extension at runtime by using monkeypatching.
+
+    The reason might be the CrateDB-specific amendments within
+    ``visit_update_14`` or even general problems with the new
+    architecture of SQLAlchemy 1.4 vs. the needs of CrateDB.
+
+    See also:
+    - https://docs.sqlalchemy.org/en/14/changelog/migration_14.html#rowproxy-is-no-longer-a-proxy-is-now-called-row-and-behaves-like-an-enhanced-named-tuple
+    - https://github.com/sqlalchemy/sqlalchemy/blob/rel_1_4_0b1/lib/sqlalchemy/cextension/resultproxy.c
+    """
```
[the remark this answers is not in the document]
amotl: The monkeypatch was removed with ed2e1e14. It was needed because there were inaccuracies when setting up fake cursors within the test cases. SQLAlchemy <1.4 might have been more forgiving on that, but >=1.4 isn't.

amotl:
A report on the notes at https://github.com/crate/crate-python/pull/391#pullrequestreview-974578917:

### 1. SQL Compilation Caching
> Special attention should be given to the new SQL Compilation Caching feature.

5f515c28a has the corresponding adjustments. The test suite succeeded, does it mean we can trust it? For further reference, an overview about this topic is at [enable caching for the given objects], with specific details about

- how to [flag custom types as "safe to cache"]
- introduction to [caching for third party dialects]
- how to [enable caching support for custom constructs]

[enable caching for the given objects]: https://docs.sqlalchemy.org/en/14/faq/performance.html#step-three-enable-caching-for-the-given-objects-and-or-seek-alternatives
[flag custom types as "safe to cache"]: https://docs.sqlalchemy.org/en/14/core/type_api.html#sqlalchemy.types.ExternalType.cache_ok
[caching for third party dialects]: https://docs.sqlalchemy.org/en/14/core/connections.html#caching-for-third-party-dialects
[enable caching support for custom constructs]: https://docs.sqlalchemy.org/en/14/core/compiler.html#enabling-caching-support-for-custom-constructs


### 2. Changelog adjustments

> Go through the list of significant changes at [What’s New in SQLAlchemy 1.4?] thoroughly and summarize/compile relevant details.

As a followup to https://github.com/crate/crate-python/pull/391#discussion_r874166450, I added 09b3173b. Let me know if you think this section about SA14 should be improved further.

[What’s New in SQLAlchemy 1.4?]: https://docs.sqlalchemy.org/en/14/changelog/migration_14.html


### 3. Test Asynchronous I/O with asyncpg

Not so fast. I created #423 to track this topic.

amotl:
Hi again,

I think the patch is ready for review and testing. We won't be running alpha/beta/rc releases, so we want to encourage the community to test the updated driver by installing it manually like:
```shell
pip install --upgrade "git+https://github.com/crate/crate-python@amo/sqlalchemy14#egg=crate[sqlalchemy]"
```

We will be happy about any kind of feedback. You can follow the further development of this patch by subscribing to the pull request for notifications.

With kind regards,
Andreas.

src/crate/client/sqlalchemy/compiler.py
```diff
@@ -74,7 +79,24 @@ def rewrite_update(clauseelement, multiparams, params):
 def crate_before_execute(conn, clauseelement, multiparams, params):
```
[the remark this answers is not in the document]
amotl: At https://github.com/sqlalchemy/sqlalchemy/discussions/5915#discussioncomment-2782664, I shared further thoughts and asked for guidance how to improve that specific topic.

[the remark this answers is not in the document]
seut: > I haven't followed Mike's suggestion yet, after I was able to make the patch work with https://github.com/crate/crate-python/commit/39d010105efa1dc5df6007b986b8d016c14e6a20. We will probably have to decide whether to drop support for SQLAlchemy 1.1 and 1.2? What are your opinions about this?

As Mike already pointed out, there are 1) not much users using these old version 2) if the do, they probably use an older version of our driver as well.
So +1 for dropping the support of SQLAlchemy < 1.3.

src/crate/client/sqlalchemy/compiler.py
```diff
@@ -71,10 +73,27 @@ def rewrite_update(clauseelement, multiparams, params):
...
+        clauseelement, multiparams, params = rewrite_update(clauseelement, multiparams, params)
+
+        if SA_VERSION >= SA_1_4:
+            if multiparams[0]:
+                params = multiparams[0][0]
+            else:
+                params = multiparams[0]
+            multiparams = []
+
+        return clauseelement, multiparams, params
```
seut, line 95: very minor: Is this return here needed? looks like if falling through it would return the same.

src/crate/client/sqlalchemy/doctests/reflection.txt
```diff
@@ -17,6 +17,14 @@ List all tables::
...
     True
 
+List all views::
+
+    >>> inspector.get_view_names()
+    []
+
+    >>> inspector.get_view_names(schema='sys')
+    []
+
```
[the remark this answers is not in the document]
seut: Why not add views to the fixtures? ;)

src/crate/client/sqlalchemy/doctests/reflection.txt
```diff
@@ -17,6 +17,14 @@ List all tables::
     >>> set(['checks', 'cluster', 'jobs', 'jobs_log']).issubset(inspector.get_table_names(schema='sys'))
     True
 
+List all views::
+
+    >>> inspector.get_view_names()
+    []
+
+    >>> inspector.get_view_names(schema='sys')
```
seut, line 25: schema `sys` will never contain any view as its an internal schema.
```suggestion
    >>> inspector.get_view_names(schema='my_schema')
```

src/crate/client/sqlalchemy/tests/array_test.py
```diff
@@ -39,7 +39,7 @@
 class SqlAlchemyArrayTypeTest(TestCase):
 
     def setUp(self):
-        self.engine = sa.create_engine('crate://')
+        self.engine = sa.create_engine('crate://cratedb.example.org/?cratedb-version=4.3.2')
```
seut, line 42: why is forcing the version here needed?

src/crate/client/sqlalchemy/tests/bulk_test.py
```diff
@@ -37,7 +37,7 @@
 class SqlAlchemyBulkTest(TestCase):
 
     def setUp(self):
-        self.engine = sa.create_engine('crate://')
+        self.engine = sa.create_engine('crate://cratedb.example.org/?cratedb-version=4.3.2')
```
seut, line 40: same here

src/crate/client/sqlalchemy/tests/insert_from_select_test.py
```diff
@@ -66,17 +68,41 @@ class CharacterArchive(Base):
         self.character_archived = CharacterArchive
         self.session = Session()
 
+    @unittest.skipIf(SA_VERSION >= SA_1_4, "Not supported on SQLAlchemy >= 1.4")
     @patch('crate.client.connection.Cursor', FakeCursor)
-    def test_insert_from_select_triggered(self):
+    def test_insert_from_select_triggered_1_0_0(self):
```
seut, line 73:
```suggestion
    def test_insert_from_select_on_1_0_0_contains_parenthesis(self):
```

src/crate/client/sqlalchemy/tests/insert_from_select_test.py
```diff
@@ -66,17 +68,41 @@ class CharacterArchive(Base):
...
+
         self.session.execute(ins)
         self.session.commit()
         self.assertSQL(
             "INSERT INTO characters_archive (name, age) (SELECT characters.name, characters.age FROM characters WHERE characters.status = ?)",
             ins
         )
+
+    @patch('crate.client.connection.Cursor', FakeCursor)
+    def test_insert_from_select_triggered_1_0_1(self):
```
seut, line 92: Is this test needed? Even on recent CrateDB versions, adding parenthesis wouldn't hurt. Look like this tests sqlalchemy and not crate specific code...

seut (commented):
Great work, thanks a lot. I've just added more minor suggestions.

src/crate/client/sqlalchemy/dialect.py
```diff
@@ -190,6 +198,13 @@ def do_rollback(self, connection):
         pass
 
     def connect(self, host=None, port=None, *args, **kwargs):
+
+        # Optionally set server version through SQLAlchemy connection string
+        # query parameter explicitly.
+        if 'cratedb-version' in kwargs:
+            forced_version = LooseVersion(kwargs.pop('cratedb-version'))
+            self.server_version_info = tuple(forced_version.version)
```
mfussenegger, line 206: Why do we need to set the cratedb version via connection string? I don't think this is something we should expose for users.

If this is required for testing, wouldn't it be possible to mock the cursor accordingly?

mfussenegger (commented):
There is a lot of code in the compiler that has no coverage. Can we either increase the amount of test or look into removing not requried stuff from it?

src/crate/client/sqlalchemy/tests/insert_from_select_test.py
amotl: The tests have already been there. They seem to specifically, and intentionally, test a change in the CrateCompiler's behavior when observing to be connected to CrateDB >= 1.0.1.

In fact, not using parentheses here is the standard behavior of SQLAlchemy. So, if we consider to drop support for CrateDB < 1.0.1, we could get rid of having to vendor and override the dialect's `visit_insert` method at all.

https://github.com/crate/crate-python/blob/02f6529d98ad75c1bb6f43f5e4ac3f81003e7f93/src/crate/client/sqlalchemy/compiler.py#L204-L219

In turn, our test suite would implictly increase code coverage because of the reduced surface of vendored code, see also https://github.com/crate/crate-python/pull/391#pullrequestreview-981282894.

src/crate/client/sqlalchemy/dialect.py
amotl: You are right, its purpose was only to cover some test cases. Exposing it for users indeed is a bad idea. I've submitted d91d088c7972 to handle that differently. I hope you like it better.

Also, when removing support for CrateDB < 1.0.1, we can get rid of this adjustment completely.

src/crate/client/sqlalchemy/tests/insert_from_select_test.py
amotl: Thanks. I've adjusted the function name in the original commit.

src/crate/client/sqlalchemy/tests/array_test.py
amotl: I needed it to satisfy some test cases which validate behavior between CrateDB < 1.0.1 vs. CrateDB >= 1.0.1.
@mfussenegger had the same question, see https://github.com/crate/crate-python/pull/391#discussion_r879126382. In order to improve the situation _how_ this is implemented, I've added d91d088c.

src/crate/client/sqlalchemy/doctests/reflection.txt
amotl: I've added 6f885e7cf34 to improve the lack of tests on this topic, which also removes inquiring VIEWs from the internal `sys` schema.

src/crate/client/sqlalchemy/compiler.py
amotl: Thanks for spotting it, that slipped through. I've removed the redundant `return` statement with the original commit which accidentally introduced it.

amotl:
Responding to @mfussenegger's question at https://github.com/crate/crate-python/pull/391#pullrequestreview-981282894:

> There is a lot of code in the compiler that has no coverage. Can we either increase the amount of test or look into removing not required stuff from it?

I think the best way forward is to remove support for older versions of CrateDB and SQLAlchemy (see also https://github.com/crate/crate-python/pull/391#discussion_r874170754) and then see what's left.

In general, the footprint of code not covered by test cases is not small because we have to vendor central compiler methods like `visit_insert`, `visit_update` and `_get_crud_params`, because only some details in them have to be amended to account for CrateDB specifics. That situation got slightly worse through recent additions of the corresponding variants taken from the SQLAlchemy 1.4 code base. I chose to do it this way deliberately because I didn't want to clutter the innards of those methods with conditional branches based on version checks.

I would prefer to work on corresponding cleanups and improvements on behalf of the next iteration, and another patchset. Otherwise, if you think the situation should be improved right now, can I humbly ask about further guidance how to do it? For example, do you think we should add the SA11/SA12 deprecation to this patchset right away?

src/crate/client/sqlalchemy/compiler.py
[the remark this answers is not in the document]
amotl: All right, let's do it. I would prefer to bring in corresponding cleanups and improvements on behalf of the next iteration, if you don't have any objections.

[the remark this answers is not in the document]
amotl: #424 implemented the update to drop support for corresponding EOL software components.

src/crate/client/sqlalchemy/tests/insert_from_select_test.py
amotl: This also became obsolete with #424.

src/crate/client/sqlalchemy/dialect.py
amotl: The adjustment became altogether obsolete with #424.

src/crate/client/sqlalchemy/compiler.py
```diff
@@ -325,3 +346,312 @@ def _get_crud_params(compiler, stmt, **kw):
...
+            set(parameters)
+            .intersection(_column_as_key(k) for k, v in stmt_parameter_tuples)
+            .difference(check_columns)
+        )
+        if check:
+            raise exc.CompileError(
+                "Unconsumed column names: %s"
+                % (", ".join("%s" % (c,) for c in check))
+            )
+    """
```
seut, line 629: The comment does not really help me in understanding the issue like why are there unconsumed column names.
Is it still valid? If so can you please comment more about the why?

src/crate/client/sqlalchemy/compiler.py
```diff
@@ -325,3 +346,312 @@ def _get_crud_params(compiler, stmt, **kw):
...
+            stmt_parameter_tuples,
+            _column_as_key,
+            values,
+            kw,
+        )
+
+    check_columns = {}
+
+    # special logic that only occurs for multi-table UPDATE
+    # statements
```
seut, line 572: I think the code speaks for itself here, I don't think the comment adds much value

amotl: This comment is part of the code which was vendored from upstream SA14.

lgtm.yml
```diff
@@ -0,0 +1,12 @@
+queries:
+
+  # Suppress some LGTM warnings.
+
+  # A module is imported with the "import" and "import from" statements.
+  # https://lgtm.com/rules/1818040193/
+  - exclude: py/import-and-import-from
```
seut, line 7: does this work? https://github.com/crate/crate-python/pull/391#issuecomment-1143438689

seut (approved):
Looks good to me. Maybe wait for @mfussenegger to have a final look at it?

lgtm.yml
amotl: No it doesn't ;[. At least, it does not work while not being merged. I hope that the additional rule will be recognized after the merge. However, it is weird because the other rule apparently seems to work.

PR #395, opened 2021-02-22
Datetime format

Aymaru:

Changing the return type from set to list on get_pk_constraint(), solves a compatibility issue from Apache Superset that doesn't show the metadata from CrateDB tables in the SQL Lab.

CrateDB types TIMESTAMP and TIMESTAMP WITHOUT TIME ZONE, are returned as python datetime objects, correctly displaying date columns in charts and SQL Lab from Apache Superset.

## Checklist

 - [x ] [CLA](https://crate.io/community/contribute/cla/) is signed

### DIFF
--- a/src/crate/client/cursor.py
+++ b/src/crate/client/cursor.py
@@ -22,6 +22,7 @@
 from .exceptions import ProgrammingError
 from distutils.version import StrictVersion
 import warnings
+from datetime import datetime
 
 BULK_INSERT_MIN_VERSION = StrictVersion("0.42.0")
 
@@ -53,7 +54,34 @@ class Cursor(object):
         self._result = self.connection.client.sql(sql, parameters,
                                                   bulk_parameters)
         if "rows" in self._result:
+            if "col_types" in self._result:
+                col_types = self._result["col_types"]
+                tmp_data = self._result["rows"]
+
+                rows_to_convert = self._get_rows_to_convert_to_date(col_types)
+                tmp_data = self._convert_dates_to_datetime(tmp_data, rows_to_convert)
+
+                self._result["rows"] = tmp_data
+
             self.rows = iter(self._result["rows"])
+
+    @staticmethod
+    def _get_rows_to_convert_to_date(col_types):
+        return [True if col_type == 11 or col_type == 15 else False for col_type in col_types]
+
+    @staticmethod
+    def _date_to_datetime(row, rows_to_convert):
+        return list(
+            map(lambda x, y:
+                datetime.fromtimestamp(float(str(x)[0:10])) if y else x,
+                row,
+                rows_to_convert))
+
+    def _convert_dates_to_datetime(self, rows, rows_to_convert):
+        return list(
+            map(lambda x:
+                self._date_to_datetime(x, rows_to_convert),
+                rows))
 
     def executemany(self, sql, seq_of_parameters):
         """
--- a/src/crate/client/sqlalchemy/dialect.py
+++ b/src/crate/client/sqlalchemy/dialect.py
@@ -43,6 +43,7 @@ TYPES_MAP = {
     "smallint": sqltypes.SmallInteger,
     "timestamp": sqltypes.TIMESTAMP,
     "timestamp with time zone": sqltypes.TIMESTAMP,
+    "timestamp without time zone": sqltypes.TIMESTAMP,
     "object": Object,
     "integer": sqltypes.Integer,
     "long": sqltypes.NUMERIC,
@@ -64,6 +65,7 @@ try:
     TYPES_MAP["smallint_array"] = ARRAY(sqltypes.SmallInteger)
     TYPES_MAP["timestamp_array"] = ARRAY(sqltypes.TIMESTAMP)
     TYPES_MAP["timestamp with time zone_array"] = ARRAY(sqltypes.TIMESTAMP)
+    TYPES_MAP["timestamp without time zone_array"] = ARRAY(sqltypes.TIMESTAMP)
     TYPES_MAP["long_array"] = ARRAY(sqltypes.NUMERIC)
     TYPES_MAP["bigint_array"] = ARRAY(sqltypes.NUMERIC)
     TYPES_MAP["double_array"] = ARRAY(sqltypes.DECIMAL)
@@ -74,7 +76,6 @@ try:
     TYPES_MAP["text_array"] = ARRAY(sqltypes.String)
 except Exception:
     pass
-
 
 log = logging.getLogger(__name__)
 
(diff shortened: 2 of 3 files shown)


amotl:
Dear @Aymaru,

thanks a stack for your contribution. We will have a look and might come back to you with some comments.

With kind regards,
Andreas.

amotl:
Dear @Aymaru,

apologies that the CI job apparently didn't work when you submitted your patch. I enabled it again (see https://github.com/crate/crate-python/pull/396#issuecomment-784370324) and just triggered it by pushing (and then removing) an empty commit to your PR.

The outcome is that the patch currently doesn't pass the tests. The spots are:

- https://github.com/crate/crate-python/runs/2003433298?check_suite_focus=true#step:5:40
- https://github.com/crate/crate-python/runs/2003433298?check_suite_focus=true#step:5:64
- https://github.com/crate/crate-python/runs/2003433298?check_suite_focus=true#step:5:138
- https://github.com/crate/crate-python/runs/2003433298?check_suite_focus=true#step:5:210
- https://github.com/crate/crate-python/runs/2003433298?check_suite_focus=true#step:5:348
- https://github.com/crate/crate-python/runs/2003433298?check_suite_focus=true#step:5:396

Will you be able to fix those?

With kind regards,
Andreas.

src/crate/client/cursor.py
```diff
@@ -53,8 +54,35 @@ def execute(self, sql, parameters=None, bulk_parameters=None):
...
+            if "col_types" in self._result:
+                col_types = self._result["col_types"]
+                tmp_data = self._result["rows"]
+
+                rows_to_convert = self._get_rows_to_convert_to_date(col_types)
+                tmp_data = self._convert_dates_to_datetime(tmp_data, rows_to_convert)
+
+                self._result["rows"] = tmp_data
+
             self.rows = iter(self._result["rows"])
```
amotl, line 66: Maybe we can come up with a more elegant and efficient implementation. When trying to follow the code, I recognize that it builds up a list of boolean flags which designates which columns to apply the datetime conversion to, right?

Based on that information, the code will then apply the conversion function on each column while mapping each row into memory through the `list()` function.

I am a bit hesitant on this and wanted to share my specific concerns:

a) The code might allocate a significant amount of memory on larger results.
b) On a very large result set, it might even run out of memory.
c) When applying the conversion, there are currently no sanity checks around which test for `NULL`- or other emptyness. I am specifically looking at `float(str(x)[0:10])` here. Also, there is no other exception handling either. We might want to improve on this aspect.

Maybe we can wrap that conversion into a generator function in order to keep the memory footprint low and, while being at it, add some more bells and whistles on the error handling side?

amotl:
Dear @Aymaru,

thank you for continuing your work on this. Please let me know if you are still able to spend time for more iterations on this or whether we should take over. I will add some more comments.

With kind regards,
Andreas.

src/crate/client/cursor.py
```diff
@@ -53,8 +54,37 @@ def execute(self, sql, parameters=None, bulk_parameters=None):
         self._result = self.connection.client.sql(sql, parameters,
                                                   bulk_parameters)
         if "rows" in self._result:
+            if "col_types" in self._result:
+                rows_to_convert = self._get_rows_to_convert_to_date(self._result["col_types"])
+                for flag in rows_to_convert:
+                    if flag:
+                        t_rows = (row for row in self._result["rows"])
```
amotl, line 61: This will probably still load the full result into memory, right?

I was thinking about swapping in the generator right in between of `self.rows = iter(self._result["rows"])`, which is then responsible for converting between data types (timestamp epoch integer to Python datetime object).

In that manner, the pre-computation using `_get_rows_to_convert_to_date()` might become obsolete.

src/crate/client/sqlalchemy/dialect.py
```diff
@@ -261,7 +262,7 @@ def get_pk_constraint(self, engine, table_name, schema=None, **kw):
 
             def result_fun(result):
                 rows = result.fetchall()
-                return set(map(lambda el: el[0], rows))
+                return list(set(map(lambda el: el[0], rows)))
```
amotl, line 265: While I don't know about the specific background why there has been a `set` used here, can you elaborate why this should now return a list?

src/crate/client/doctests/client.txt
```diff
@@ -179,7 +179,7 @@ supported, all other fields are 'None'::
     >>> result = cursor.fetchone()
     >>> pprint(result)
     ['Aldebaran',
-     1373932800000,
+     datetime.datetime(2013, 7, 15, 18, 0),
```
amotl, line 182: This currently raises an exception on CI, see [1]. The current code expects `datetime.datetime(2013, 7, 16, 0, 0)` here.

[1] https://github.com/crate/crate-python/runs/2114206441?check_suite_focus=true#step:5:388

amotl:
Dear @Aymaru,

first things first: As I noticed recent activity from you on this patch again, I wanted to thank you for keeping up to it.

Further, I wanted to take the chance to tell you that we recently fixed some flaws within the test suite. Besides it now also works on macOS, more importantly, it received different fixes where the outcome of the software tests was indeterministic before. So, you might want to rebase your branch on top of the current master branch in order to also receive those improvements.

With kind regards,
Andreas.

src/crate/client/sqlalchemy/dialect.py
amotl: Dear @Aymaru,

apologies for the late reply. We've split off this amendment into #426. For you, it apparently solved this issue:

> Changing the return type from set to list on get_pk_constraint(), solves a compatibility issue from Apache Superset that doesn't show the metadata from CrateDB tables in the SQL Lab.

Maybe due to the upgrade to SQLAlchemy 1.4, and maybe other changes in Apache Superset, this is not needed anymore?

With kind regards,
Andreas.

amotl:
Dear @Aymaru,

the main aspects of your patch have been converged into #442, #445, and #447. With #426, we evaluated that this specific detail creates a regression when used with Apache Superset.

With kind regards,
Andreas.

amotl:
Dear @hlcianfagna and @Aymaru,

we would like to thank you once again for your excellent improvements on this matter. It looks like Hernan's recent improvements made it into the upstream codebase of Apache Superset, so we may consider this issue as resolved?

@Aymaru: Can you verify if everything you aimed at works well with CrateDB now, when using a recent version of Apache Superset?

With kind regards,
Andreas.

amotl:
Hi again. Based on those improvements, and that we didn't hear about any other flaws recently, we are closing this PR, in the assumption that everything works well in this regard.

- https://github.com/apache/superset/pull/27567
- https://github.com/crate/sqlalchemy-cratedb/pull/22
- https://github.com/crate/sqlalchemy-cratedb/pull/25
- https://github.com/apache/superset/pull/29243

Please let us know if you think anything is still wrong, best raised on the issue tracker of [sqlalchemy-cratedb]. Thank you again for your efforts on this!

[sqlalchemy-cratedb]: https://github.com/crate/sqlalchemy-cratedb

PR #385, opened 2020-10-28
Verify SSL as True by default

autophagy:

Issue #375, opened 2020-09-09
https.txt tests fail


chaudum:
Instead of an SSL exception the test fails when trying to establish a connection

```
Failure in test /home/christian/sandbox/crate/crate-python/src/crate/client/doctests/https.txt
Failed doctest test for https.txt
  File "/home/christian/sandbox/crate/crate-python/src/crate/client/doctests/https.txt", line 0

----------------------------------------------------------------------
File "/home/christian/sandbox/crate/crate-python/src/crate/client/doctests/https.txt", line 80, in https.txt
Failed example:
    client.server_infos(crate_host)
Expected:
    Traceback (most recent call last):
    ...
    crate.client.exceptions.ConnectionError: Server not available, exception: ...[SSL: ...
Got:
    Traceback (most recent call last):
      File "/usr/lib/python3.8/doctest.py", line 1336, in __run
        exec(compile(example.source, filename, "single",
      File "<doctest https.txt[11]>", line 1, in <module>
        client.server_infos(crate_host)
      File "/home/christian/sandbox/crate/crate-python/src/crate/client/http.py", line 340, in server_infos
        response = self._request('GET', '/', server=server)
      File "/home/christian/sandbox/crate/crate-python/src/crate/client/http.py", line 439, in _request
        raise ConnectionError(
    crate.client.exceptions.ConnectionError: Server not available, exception: HTTPSConnectionPool(host='localhost', port=65534): Max retries exceeded with url: / (Caused by NewConnectionError('<urllib3.connection.HTTPSConnection object at 0x7f0779d09730>: Failed to establish a new connection: [Errno 110] Connection timed out'))

```

amotl:
This error is tripping again on CI, see [1].

[1] https://github.com/crate/crate-python/runs/2115670144?check_suite_focus=true#step:5:156

amotl:
Just to share an observation about this....

When running `src/crate/client/doctests/https.txt` in isolation, this warning appears at the end of the test run:
```python
sys:1: ResourceWarning: unclosed <socket.socket fd=4, family=AddressFamily.AF_INET, type=SocketKind.SOCK_STREAM, proto=0, laddr=('127.0.0.1', 65534)>
```

Others are observing the same thing when HTTP requests are issued within a unittest environment, see [1,2]. I tried different countermeasures but haven't been able to resolve it.

[1] https://stackoverflow.com/questions/48160728/resourcewarning-unclosed-socket-in-python-3-unit-test
[...]

PR #397, opened 2021-03-15
Refresh the development enviroment, care about missed "verify_ssl_cert=True" and stabilize/fix doctests in "https.txt"

amotl:
Hi there,

when using Python 3.9, setting up the development environment croaked. Because `bootstrap.py` was a workaround in pre-`pip` times [1], and to modernize this repository, this patch completely gets rid of it. On the other hand, it pins to appropriate package versions to still be compatible with Python 3.5 through Python 3.9, where the former setup didn't satisfy these requirements anymore.

Along the lines, the patch fixes the doctests in `https.txt` in order to improve the situation with #375. For doing this appropriately, it takes care about a missed `verify_ssl_cert=True` setting on the HTTP client as a followup to #385.

We need this to be able to run #400 successfully on CI.

With kind regards,
Andreas.

[1] https://community.plone.org/t/not-using-bootstrap-py-as-default/620


amotl:
> slightly improves the test layer setup machinery as an attempt to improve the situation with #375

We are not there yet. Because this error is still there, this patch switches to `fail-fast: false` on the test matrix in order to at least run it on all available Python versions instead of failing the complete suite immediately after tripping into the `https.txt` problem.

However, because of [1], we need this patch to be integrated before being able to run #400 on CI.

[1] https://github.com/crate/crate-python/pull/400/checks?check_run_id=2133451691#step:4:575

mfussenegger:
It's a bit hard to review as this PR does several things. Any chance to break this up into multiple smaller ones?

amotl:
> It's a bit hard to review as this PR does several things. Any chance to break this up into multiple smaller ones?

I can do it. But I will probably not be able to make CI succeed on individual chunks because currently, it is broken due to both the environment being outdated because of issues with `buildout` and the nasty `https.txt` woes on the other side of the story. However, I can try to break out 06067a2, 4eeff6f, a63e702 and d2fb2b5 into a different patch, or, on the other hand, hand in 1081140 separately.

amotl:
Hi @mfussenegger and @seut,

I removed some of the commits here in order to boil it down to the minimum required things to make CI succeed again. I will submit a separate patch including them.

With kind regards,
Andreas.

DEVELOP.rst
```diff
@@ -1,46 +1,57 @@
...
 This project uses buildout_ to set up the development environment.
 
-To start things off, run::
+To start things off, create a Python virtualenv and install buildout::
 
-    $ python bootstrap.py
+    python3 -m venv .venv
+    source .venv/bin/activate
+
+    # Workaround for Python 3.9
```
seut, line 15:
```suggestion
    # Workaround for Python 3.5
```

seut (approved):
Looks good otherwise.

DEVELOP.rst
amotl: Thanks for spotting this.

mfussenegger (approved):

PR #400, opened 2021-03-17
Support credentials and SSL with SQLAlchemy via HTTP/DB URIs

### LINKED ISSUE #399: Support HTTPS URLs when specifying a SqlAlchemy URI
There are two ways to specify a server to connect when using SqlAlchemy. The official one recommended by the [crate docs](https://crate.io/docs/python/en/latest/sqlalchemy.html#using-sqlalchemy) is to use the `connect_args` like this:

```python
engine = sa.create_engine(
     'crate://',
     connect_args={
         'servers': ['198.51.100.1:4200', '198.51.100.2:4200'],
         'verify_ssl_cert': True,
         'ca_cert': '<PATH_TO_CA_CERT>',
     }
)
```

There is a simpler way to do this, by using just the URI:
```
crate://user:pass@host
```
however, in this case the crate driver defaults to HTTP, and there is no way to specify that it is a HTTPS connection. This should be supported for ease of use.

amotl:
Hi there,

in order to resolve #399, this patch improves some bits when connecting to CrateDB, specifically in scenarios when preferring to connect using a SQLAlchemy DB URI instead of separate `connect_args`. This is convenient for programs which can be configured using a single DB URI as database connection option, like Apache Superset.

- 064e341a4: Accept credentials within the HTTP URI. Example: `http://foo:bar@cratedb.example.org:4200/`
- 52e715d0: Enable SSL using SQLAlchemy DB URI. Example: `crate://cratedb.example.org/?ssl=true`

With kind regards,
Andreas.

P.S.: 30fc28e, 93f33b96 and 160aa1e9 are only maintenance commits without feature significance.

### DIFF
--- a/CHANGES.txt
+++ b/CHANGES.txt
@@ -13,6 +13,9 @@ Unreleased
 
 - Adjusted connect arguments to accept credentials within the HTTP URI.
 
+- Added support for enabling SSL using SQLAlchemy DB URI with parameter
+  ``?ssl=true``.
+
 2020/09/28 0.26.0
 =================
 
--- a/CHANGES.txt
+++ b/CHANGES.txt
@@ -11,6 +11,8 @@ Unreleased
   HTTP by default. Previously, this setting defaulted to false. This setting
   can be changed via the ``verify_ssl_cert`` connection parameter.
 
+- Adjusted connect arguments to accept credentials within the HTTP URI.
+
 2020/09/28 0.26.0
 =================
 
--- a/src/crate/client/sqlalchemy/dialect.py
+++ b/src/crate/client/sqlalchemy/dialect.py
@@ -25,6 +25,7 @@ from datetime import datetime, date
 from sqlalchemy import types as sqltypes
 from sqlalchemy.engine import default, reflection
 from sqlalchemy.sql import functions
+from sqlalchemy.util import asbool, to_list
 
 from .compiler import (
     CrateCompiler,
@@ -194,8 +195,12 @@ class CrateDialect(default.DefaultDialect):
             server = '{0}:{1}'.format(host, port or '4200')
         if 'servers' in kwargs:
             server = kwargs.pop('servers')
-        if server:
-            return self.dbapi.connect(servers=server, **kwargs)
+        servers = to_list(server)
+        if servers:
+            use_ssl = asbool(kwargs.pop("ssl", False))
+            if use_ssl:
+                servers = ["https://" + server for server in servers]
+            return self.dbapi.connect(servers=servers, **kwargs)
         return self.dbapi.connect(**kwargs)
 
     def _get_default_schema_name(self, connection):
--- a/.github/workflows/tests.yml
+++ b/.github/workflows/tests.yml
@@ -18,7 +18,7 @@ jobs:
       matrix:
         os: [ubuntu-latest, macos-latest]
         python-version: [3.5, 3.6, 3.7, 3.8, 3.9]
-        cratedb-version: ['4.5.0']
+        cratedb-version: ['4.7.1']
         sqla-version: ['1.1.18', '1.2.19', '1.3.23']
       fail-fast: true
 
--- a/docs/appendices/data-types.rst
+++ b/docs/appendices/data-types.rst
@@ -48,19 +48,19 @@ CrateDB       Python
 
 __ https://crate.io/docs/crate/reference/en/latest/general/ddl/data-types.html#boolean
 __ https://docs.python.org/3/library/stdtypes.html#boolean-values
-__ https://crate.io/docs/crate/reference/en/latest/general/ddl/data-types.html#character-data-types
+__ https://crate.io/docs/crate/reference/en/latest/general/ddl/data-types.html#character-data
 __ https://docs.python.org/3/library/stdtypes.html#str
-__ https://crate.io/docs/crate/reference/en/latest/general/ddl/data-types.html#numeric-types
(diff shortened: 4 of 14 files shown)


CHANGES.txt
```diff
@@ -13,6 +13,8 @@ Unreleased
 
 - Adjusted connect arguments to accept credentials within the HTTP URI.
 
+- Adjusted connect arguments to enable SSL using SQLAlchemy DB URI.
```
mfussenegger, line 16: Maybe rephrase this to make it more end-user comprehensible

src/crate/client/sqlalchemy/dialect.py
```diff
@@ -194,8 +195,16 @@ def connect(self, host=None, port=None, *args, **kwargs):
             server = '{0}:{1}'.format(host, port or '4200')
         if 'servers' in kwargs:
             server = kwargs.pop('servers')
-        if server:
-            return self.dbapi.connect(servers=server, **kwargs)
+        servers = to_list(server)
+        if servers:
+            # Evaluate "ssl" connection URI query parameter.
```
mfussenegger, line 200:
```suggestion
```
Just repeats what the code is already stating without extra information

src/crate/client/sqlalchemy/dialect.py
```diff
@@ -194,8 +195,16 @@ def connect(self, host=None, port=None, *args, **kwargs):
...
+            # TODO: Evaluate more parameters like `ssl_ca`, `ssl_key`, `ssl_cert`,
+            #       `ssl_capath` and `ssl_cipher`.
```
mfussenegger, line 202:
```suggestion
```

TODOs tend to stick around for way longer then they should. I think this should be tracked in an issue or just remove it. If it is important enough it will come up again.

src/crate/client/sqlalchemy/dialect.py
```diff
@@ -194,8 +195,16 @@ def connect(self, host=None, port=None, *args, **kwargs):
...
+            if "ssl" in kwargs:
+                use_ssl = asbool(kwargs.pop("ssl"))
+                if use_ssl:
```
mfussenegger, line 205:
```suggestion
              use_ssl = asbool(kwargs.pop("ssl"))
              if use_ssl:
```

Looks like this should work too

mfussenegger (commented):
> P.S.: https://github.com/crate/crate-python/commit/30fc28e1008e572de8baa5144e416f660c530fe6, https://github.com/crate/crate-python/commit/93f33b968cbff1b2a3816b4492ad78abf1a7ad85 and https://github.com/crate/crate-python/commit/160aa1e95a3df940b6871c431d25a94864f12c86 are only maintenance commits without feature significance

Would be cool to split maintenance stuff and features into separate PRs in the future. It is much easier to review that way. Github doesn't support adding suggestions when looking at an individual commit.

CHANGES.txt
amotl: Good idea. I tried to rephrase it with 9984cc4826.

src/crate/client/sqlalchemy/dialect.py
amotl: Also removed with 9984cc4826.

src/crate/client/sqlalchemy/dialect.py
amotl: So true. 9984cc4826 removes the inline comment. Thanks.

src/crate/client/sqlalchemy/dialect.py
amotl: I don't think it will work but raise an exception instead, when `ssl` is not available within the parameter dictionary. Am I missing something?
```
>>> a = {}
>>> a.pop("b")
  File "<stdin>", line 1, in <module>
KeyError: 'b'
```

mfussenegger: Right, would have to use `kwargs.pop("ssl", False)` or with `None`

CHANGES.txt
```diff
@@ -11,6 +11,10 @@ Unreleased
   HTTP by default. Previously, this setting defaulted to false. This setting
   can be changed via the ``verify_ssl_cert`` connection parameter.
 
+- Adjusted connect arguments to accept credentials within the HTTP URI.
+
+- Optionally enable SSL using SQLAlchemy DB URI with parameter ``?ssl=true``.
```
mfussenegger, line 16:
```suggestion
- Added support for enabling SSL using SQLAlchemy DB URI with parameter ``?ssl=true``.
```

mfussenegger (commented):
The changes should probably also get documented somewhere in https://github.com/crate/crate-python/blob/1b1136a3b919c86d11fb5d2ab9adc5db3dd7fac6/docs/sqlalchemy.rst#L47-L129 

Otherwise I think this looks good

src/crate/client/sqlalchemy/dialect.py
amotl: Ah, right. Addressed with f1b22382f5.

CHANGES.txt
amotl: Thanks. Your suggestion has also been implemented with f1b22382f5.

amotl:
> The changes should probably also get documented somewhere.

Thank you for your attention. I've just added 8fa85a5eb6 and c08ecd142, intending to improve the corresponding documentation about the new capabilities to connect by URI.

mfussenegger (approved):

PR #401, opened 2021-03-24
Improve test suite

amotl:
Hi there,

based on this patch, it becomes easier to selectively run parts of the test suite. This is important to me because the whole test suite is currently only able to be invoked exclusively on Linux.

With kind regards,
Andreas.

### DIFF
--- a/src/crate/client/tests.py
+++ b/src/crate/client/tests.py
@@ -23,6 +23,7 @@ from __future__ import absolute_import
 
 import json
 import os
+import socket
 import unittest
 import doctest
 from pprint import pprint
@@ -32,6 +33,8 @@ import ssl
 import time
 import threading
 import logging
+
+import stopit
 
 from crate.testing.layer import CrateLayer
 from crate.testing.tests import crate_path, docs_path
@@ -258,7 +261,7 @@ class HttpsTestServerLayer:
         thread = threading.Thread(target=self.serve_forever)
         thread.daemon = True  # quit interpreter when only thread exists
         thread.start()
-        time.sleep(0.5)
+        self.waitForServer()
 
     def serve_forever(self):
         print("listening on", self.HOST, self.PORT)
@@ -267,6 +270,29 @@ class HttpsTestServerLayer:
 
     def tearDown(self):
         self.server.shutdown()
+
+    def isUp(self):
+        """
+        Test if a host is up.
+        """
+        s = socket.socket(socket.AF_INET, socket.SOCK_STREAM)
+        ex = s.connect_ex((self.HOST, self.PORT))
+        s.close()
+        return ex == 0
+
+    def waitForServer(self, timeout=5):
+        """
+        Wait for the host to be available.
+        """
+        with stopit.ThreadingTimeout(timeout) as to_ctx_mgr:
+            while True:
+                if self.isUp():
+                    break
+                time.sleep(0.001)
+
+        if not to_ctx_mgr:
+            raise TimeoutError("Could not properly start embedded webserver "
+                               "within {} seconds".format(timeout))
 
 
 def setUpWithHttps(test):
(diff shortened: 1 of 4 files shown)


seut:
@amotl what about the test failures?

seut (approved):
👍

seut:
Ok got it, test failures should be solved by #397.

src/crate/client/tests.py
```diff
@@ -108,15 +108,22 @@ def setUpMocked(test):
...
+    global crate_layer
+
+    if crate_layer is None:
+        crate_layer = CrateLayer('crate',
+                                 crate_home=crate_path(),
+                                 port=crate_port,
+                                 host=local,
+                                 transport_port=crate_transport_port,
+                                 settings=settings)
+    return crate_layer
```
mfussenegger, line 126: I don't really understand how this improves the situation?

src/crate/client/tests.py
```diff
@@ -248,7 +255,7 @@ def setUp(self):
         thread = threading.Thread(target=self.serve_forever)
         thread.daemon = True  # quit interpreter when only thread exists
         thread.start()
-        time.sleep(1)
+        time.sleep(0.5)
```
mfussenegger, line 258: This seems a bit arbitrary. If we wanted to make sure the server is ready I think we should use some kind of wait/try/backoff pattern like we do in `layer.py` with `wait_for_http_url`

src/crate/client/tests.py
amotl: Because then, I can selectively manually disable some tests within the backplane `tests.py` and make the test suite still partially run on macOS. That is an alternative to the other patch at #386.

Based on this improvement, I might submit another patch which can enable/disable specific tests selectively by providing some environment variables.

src/crate/client/tests.py
amotl: I can look into also adding this here.

amotl: I've just improved this on behalf of 6472212.

src/crate/client/tests.py
mfussenegger: Just to make sure I understand this right: You want to be able to run individual tests which do not need the test layer without them triggering the test layer startup? (And the test selection is done as usual with the test runner) ?

If so :+1:  
but maybe you could add this information as comment around this code?

amotl: I just added 0e59669c9 to describe the use case a bit more.

> You want to be able to run individual tests which do not need the test layer without them triggering the test layer startup?

Exactly.

> And the test selection is done as usual with the test runner.

I am not sure if this is possible yet, because the tests in question are doctests. I tried some variants, but I haven't been able to select specific ones based on e.g. `--tests-pattern` yet.

src/crate/client/tests.py
```diff
@@ -108,15 +111,33 @@ def setUpMocked(test):
...
+
+def ensure_cratedb_layer():
+    """
+    In order to skip individual tests by manually disabling them within
+    `def test_suite()`, it is crucial make the test layer not run on each
+    and every occasion. So, things like this will be possible::
+
+        ./bin/test -vvvv --ignore_dir=testing
+
+    TODO: Through a subsequent patch, the possibility to individually
```
mfussenegger, line 127: We could use the logic from here https://github.com/crate/crate/commit/414cd8339b216478a7a06b10e4c6353dba7cfc74

mfussenegger (approved):

src/crate/client/tests.py
amotl: Oh this is great. That is exactly what I had in mind here. With #386, the need for having this immediately got a lower priority for me, but I will come back to this as soon as we will need it.

amotl: I added your suggestion on behalf of 3335961f29. Thanks again!

PR #402, opened 2021-03-24
Close urllib3 pool before swapping the server into the "inactive" list

amotl:
Hi there,

while working on the test suite, I found a glitch, maybe unnoticed more often than not, where active connections have not been closed before removing them from the list of active servers.

With kind regards,
Andreas.


seut (approved):
👍 good catch

PR #403, opened 2021-03-24
Some adjustments re. switchover to default certificate verification

amotl:
Hi there,

there were some more leftovers, where the switchover to "make SSL certificate verification a default" (`verify_ssl_cert=True`, #385) missed out. This patch accounts for that.

With kind regards,
Andreas.


seut (approved):
👍

mfussenegger (approved):

PR #424, opened 2022-05-24
Drop support for EOL software components

amotl:
As a precursor to #391, this patch drops support for software components considered to be EOL.

- Drop support for SQLAlchemy 1.1 and 1.2.
- Drop support for CrateDB < 2.0.0.
- Drop support for Python 3.4, 3.5, 3.6.


mfussenegger (approved):
Would be nice to put the change that drops support and the related documentation update into the same commit. Otherwise lgtm

seut (commented):
Can't the `compiler.CrateCompiler.visit_insert` method be removed  now as supported CrateDB versions do support insert-from-select without parenthesis?

amotl:
Hi again,

I've amended the commits to include the corresponding updates to the documentation and also added 08ed75c784, following up on the comment

> For example, when dropping support for CrateDB <= 1.0.0, we could get rid of vendoring the `visit_insert` compiler method completely. -- https://github.com/crate/crate-python/pull/391#issuecomment-1135753136

Thank you for your attention. 🙇‍♂️

With kind regards,
Andreas.

PR #426, opened 2022-06-17
Improve compatibility with Apache Superset by showing CrateDB table metadata in the SQL Lab editor

amotl:
Hi,

this patch contributed by @Aymaru at #395 aims to improve compatibility with Apache Superset by returning a list instead of a set from ``get_pk_constraint``. Otherwise, Apache Superset would not show CrateDB table metadata in the SQL Lab editor.

With kind regards,
Andreas.

P.S.: The patch has not been verified together with Apache Superset yet. Maybe you can have a look, @rafaelasantana, @hammerhead or @proddata?

Installing the code from the corresponding development branch conveniently would work like this:
```
pip install --upgrade "git+https://github.com/crate/crate-python@amo/fix-apache-superset-metadata#egg=crate[sqlalchemy]"
```


seut (commented):
Looks good to me, but would be nice to verify that this solves the issue first.
Is the pgsql dialect also returning a list here? (if so I'd assume it as verified :)

amotl:
With this patch, we are observing a regression. Where the primary key information was able to be inspected beforehand, with the patch it does not.

You can see this on those screenshots.

### Before
image

### After
image

amotl:
Hm. @proddata pointed out that the [documentation of `get_pk_constraint`](https://docs.sqlalchemy.org/en/14/core/internals.html#sqlalchemy.engine.default.DefaultDialect.get_pk_constraint) clearly states:

> Given a [Connection](https://docs.sqlalchemy.org/en/14/core/connections.html#sqlalchemy.engine.Connection), a string `table_name`, and an optional string `schema`, return primary key information as a dictionary with these keys:
> 
> - **constrained_columns**: a list of column names that make up the primary key.
> - **name**: optional name of the primary key constraint.

So, I figure that returning `constrained_columns` as a list should cause no troubles. Instead, it should be compliant with the specification even better.

Why would it stop working with Apache Superset then?

amotl:
Dear @Aymaru,

can I humbly ask you to have a look at this patch again? I am wondering if I **correctly** captured the gist of _that part_ of your patch, which was originally submitted at #395.

Maybe you can observe a flaw here, or maybe you have another explanation for the behavior of Apache Superset on this matter. Any help or support will be greatly appreciated.

With kind regards,
Andreas.

amotl:
I've probed this once more to make sure it is the correct outcome.

### Before
Without this patch, table metadata is displayed.
image image

### After
When running with this patch, table metadata stops being displayed.
image

amotl:
Dear @Aymaru,

based on my findings above, we will not bring in this part of the patch #395 you submitted the other day.

> Right now, we are actually observing the opposite behavior with this patch: It works without, but doesn't when applied.
> Maybe Apache Superset has advanced in this area in the meanwhile?

I will have a look at the other parts of your PR on behalf of a different patch.

With kind regards,
Andreas.

PR #437, opened 2022-07-18
Followup: Return TIMESTAMP columns as native Python datetime objects

amotl:
Dear @Aymaru,

apologies for the late reply.

On the patch you submitted at crate/crate-python#395 the other day, I recently exercised crate/crate-python#426 separately. Your other improvements from there have been converged into this very patch, which has been slightly cleaned up to reflect the changes without the gist of crate/crate-python#426 and other spurious commits introduced by merging from the master branch.

I've tried to keep your original commits for now, but I might finally squash them together while working on the patch. The next steps are rebasing upon master and adding eventual fixup commits.

Feel free to also add additional comments or suggestions, specifically if you can spot a place where I missed to reflect the improvements from your original patch crate/crate-python#395 correctly.

With kind regards,
Andreas.

---

#### Backlog
- [ ] Also convert timestamps within ARRAYs. -- https://github.com/crate/crate-python/pull/437#discussion_r930822457
- [ ] Enable this functionality only with a feature flag.

### DIFF
--- a/src/crate/client/cursor.py
+++ b/src/crate/client/cursor.py
@@ -87,7 +87,7 @@ class Cursor(object):
                 break
 
             if flag and value is not None:
-                value = datetime.fromtimestamp(value / 1e3)
+                value = datetime.utcfromtimestamp(value / 1e3)
 
             yield value
 
--- a/src/crate/client/doctests/client.txt
+++ b/src/crate/client/doctests/client.txt
@@ -204,11 +204,6 @@ fetchall()
 Cursor Description
 ==================
 
-.. hidden: Set time zone to UTC to make naive datetime objects deterministic.
-
-    >>> previous_timezone = os.environ.get("TZ")
-    >>> os.environ["TZ"] = "UTC"
-
 The ``description`` property of the cursor returns a sequence of 7-item
 sequences containing the column name as first parameter. Just the name field is
 supported, all other fields are 'None'::
@@ -243,10 +238,6 @@ supported, all other fields are 'None'::
      ('position', None, None, None, None, None, None),
      ('description', None, None, None, None, None, None),
      ('details', None, None, None, None, None, None))
-
-.. hidden: Restore time zone
-
-    >>> if previous_timezone: os.environ["TZ"] = previous_timezone
 
 
 Closing the Cursor


src/crate/client/cursor.py
```diff
@@ -49,8 +50,45 @@ def execute(self, sql, parameters=None, bulk_parameters=None):
...
+        rows_to_convert = [True if col_type == 11 or col_type == 15 else False for col_type in
+                           self._result["col_types"]]
+        for row in self._result["rows"]:
+            gen_flags = (flag for flag in rows_to_convert)
+            yield [t_row for t_row in self._transform_date_columns(row, gen_flags)]
+
+    @staticmethod
+    def _transform_date_columns(row, gen_flags):
+        """
+        Generates iterates over each value from a row and converts timestamps to pandas TIMESTAMP
```
matriv, line 76: Can we do something more elegant with [lambdas](https://www.geeksforgeeks.org/python-iterating-with-python-lambda/ ) ? (I have basic python knowledge though).

amotl:
After mitigating some errors on CI, this is the last one happening. Is it related to now using the HTTP API endpoint `/_sql?types` instead of `/_sql`?

```
Failure in test /Users/amo/dev/crate/sources/crate-python/src/crate/client/doctests/http.txt
Failed doctest test for http.txt
  File "/Users/amo/dev/crate/sources/crate-python/src/crate/client/doctests/http.txt", line 0

----------------------------------------------------------------------
File "/Users/amo/dev/crate/sources/crate-python/src/crate/client/doctests/http.txt", line 204, in http.txt
Failed example:
    print(trace)
Expected:
    TRACE: ... mismatched input 'error' expecting {<EOF>, ...
    at io.crate...
Got:
    TRACE: None
```

https://github.com/crate/crate-python/blob/e23d651b73bd50acbfbbafc2178e6e98c04c045a/src/crate/client/doctests/http.txt#L193-L205

src/crate/client/cursor.py
amotl: I think using the generator here is fine. Let me know if you think otherwise.

src/crate/client/cursor.py
```diff
@@ -49,8 +50,45 @@ def execute(self, sql, parameters=None, bulk_parameters=None):
...
+            try:
+                flag = next(gen_flags)
+            except StopIteration:
+                break
+
+            if not flag or value is None:
+                yield value
+            else:
+                if value < 0:
+                    yield None
```
amotl, line 88: Not sure about this guard?

src/crate/client/sqlalchemy/doctests/itests.txt
```diff
@@ -80,9 +80,6 @@ Date should have been set at the insert due to default value via python method::
     >>> dt.day == now.day
     True
 
-    >>> (now - location.datetime_tz).seconds < 4
-    True
-
```
amotl, line 85: Did this test case protect against anything significant? I can not make any sense of it.

src/crate/client/cursor.py
amotl: When looking at this again, specifically after cleaning up a bit with 6522e430284, I tend to agree with you. It is probably enough that `_transform_result_types`, which iterates all records, is a generator. On the other hand, `_transform_datetime_columns`, which iterates all columns per record, does not necessarily have to be a generator.

I will keep the original patch unchanged on this matter yet, and will await the official review.

src/crate/client/cursor.py
```diff
@@ -49,9 +50,47 @@ def execute(self, sql, parameters=None, bulk_parameters=None):
...
             self.rows = iter(self._result["rows"])
 
+    def _transform_result_types(self):
+        """
+        Generate row items with column values converted to their corresponding
+        native Python types, based on information from `col_types`.
+
+        Currently, only converting to native `datetime` objects is implemented.
+        """
+        datetime_column_types = [11, 15]
```
amotl, line 70: Does someone have a resource at hand where those column code types are enumerated? The table at [^1] coincidentally lists `_timestamp without time zone` as `1115`, but here the code is apparently expecting two-digit integer numbers.

I will be happy to receive further pointers for better educating myself on this topic.

[^1]: https://crate.io/docs/crate/reference/en/5.0/interfaces/postgres.html#pg-type

mfussenegger: https://crate.io/docs/crate/reference/en/5.0/interfaces/http.html#column-types


The code here will probably also need to handle arrays of timestamps and timestamps contained within objects.

amotl: Thanks. I will add the conversion for container structures as well.

Reading up on your reference, I am asking myself whether it is appropriate that `type=15` (Unchecked object) is handled here as well?

src/crate/client/cursor.py
```diff
@@ -49,9 +50,47 @@ def execute(self, sql, parameters=None, bulk_parameters=None):
...
+        Convert all designated columns to native Python `datetime` objects.
+        """
+        for value in row:
+            try:
+                flag = next(column_flags)
+            except StopIteration:
+                break
+
+            if flag and value is not None:
+                value = datetime.fromtimestamp(value / 1e3)
```
mfussenegger, line 90: I think using `utcfromtimestamp` is the less error prone option here.

There are two scenarios:

The user stores UTC on the server:

- `fromtimestamp` would be wrong, or at least confusing as it would convert to localtime but without attaching a timezone info.
- `utcfromtimestamp` would be correct

The user stores local timestamps on the server. Unless they also store the timezone in a separate field this is error prone but

- `utcfromtimestamp` would at least preserve the value used on the server
- `fromtimestamp` would be wrong, because it would apply _another_ offset.


It's important that we don't attach a `tzinfo` given that we don't have one.

mfussenegger: Another option could be to to let users  configure the timezone on the client (with default UTC) and always create timezone aware objects.

The default could turn out to be wrong, but at least the user has an easy option to fix it, and it's easier to spot a incorrectly attached tzone then it is to figure out that a conversion happened and why/where some conversion happend.
It's also way more convenient/less error prone to continue working with tz aware datetimes.

amotl:
I am closing this in favor of other patches / issues, where the corresponding details have been converged into.
- #442
- #445
- #447

PR #438, opened 2022-07-18
Tests: Also use `datetime` column within `locations` test fixtures

amotl:
Hi,

this patch will be good to have before #437 in order to better outline specific changes to `date` vs. `datetime` return values.

All the values in the `datetime` column have been filled with the same value as in the corresponding `date` column. No other values have been changed.

With kind regards,
Andreas.


src/crate/client/doctests/client.txt
```diff
@@ -213,7 +213,7 @@ supported, all other fields are 'None'::
     >>> pprint(result)
     ['Aldebaran',
      1373932800000,
-     None,
+     1373932800000,
```
matriv, line 216: maybe you can add also time info, down to millis?

amotl: Sure. a7fa1ee55f brings in a corresponding improvement.

matriv: As discussed, after that we need to add a column of `DATE` to test table `locations`, and we need also to add functionality to the client to truncate to start of day when using python date type, even if the table column (result) is a complete timestamp with time info.

matriv (approved):
LGTM, added a followup task/improvement as discussed.

mfussenegger (approved):

PR #440, opened 2022-07-21
Adjust usage of `utcnow()` and `utcfromtimestamp()` for Python 3

amotl:
### Motivation
This patch further sets the stage for #437.

### About

On the transition from Python 2 to Python 3, this subtle adjustment sometimes slips through. It is needed for properly working with the `now()` variants of native Python `date` and `datetime` objects.

- Instead of `datetime.utcnow()`, use `datetime.now(timezone.utc)`
- Instead of `utcfromtimestamp()`, use `fromtimestamp(..., tz=timezone.utc)`

See also: https://blog.ganssle.io/articles/2019/11/utcnow.html

Please note that this patch still reflects the fact that only timezone-unaware (naive) datetime objects are handled yet. In order to compensate that, a number of `.replace(tzinfo=None)` calls have been added along the lines.

### DIFF
--- a/src/crate/client/sqlalchemy/dialect.py
+++ b/src/crate/client/sqlalchemy/dialect.py
@@ -20,7 +20,7 @@
 # software solely pursuant to the terms of the relevant commercial agreement.
 
 import logging
-from datetime import datetime, date
+from datetime import datetime, date, timezone
 
 from sqlalchemy import types as sqltypes
 from sqlalchemy.engine import default, reflection
@@ -90,7 +90,8 @@ class Date(sqltypes.Date):
             if not value:
                 return
             try:
-                return datetime.utcfromtimestamp(value / 1e3).date()
+                return datetime.fromtimestamp(value / 1e3, tz=timezone.utc) \
+                    .replace(tzinfo=None).date()
             except TypeError:
                 pass
 
@@ -129,7 +130,8 @@ class DateTime(sqltypes.DateTime):
             if not value:
                 return
             try:
-                return datetime.utcfromtimestamp(value / 1e3)
+                return datetime.fromtimestamp(value / 1e3, tz=timezone.utc) \
+                    .replace(tzinfo=None)
             except TypeError:
                 pass
 
--- a/src/crate/client/sqlalchemy/doctests/itests.txt
+++ b/src/crate/client/sqlalchemy/doctests/itests.txt
@@ -65,10 +65,10 @@ Retrieve the location from the database::
     >>> location.name
     'Earth'
 
-Date should have been set at the insert due to default value via python method::
+Date should have been set at the insert due to default value via Python method::
 
-    >>> from datetime import datetime
-    >>> now = datetime.utcnow()
+    >>> from datetime import datetime, timezone
+    >>> now = datetime.now(timezone.utc).replace(tzinfo=None)
     >>> dt = location.date
 
     >>> dt.year == now.year
@@ -78,9 +78,6 @@ Date should have been set at the insert due to default value via python method::
     True
 
     >>> dt.day == now.day
-    True
-
-    >>> (now - location.datetime_tz).seconds < 4
     True
 
 Verify the return type of date and datetime::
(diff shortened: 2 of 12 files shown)


src/crate/client/tests.py
```diff
@@ -56,6 +55,7 @@
 )
 from .sqlalchemy.tests import test_suite as sqlalchemy_test_suite
 from .sqlalchemy.types import ObjectArray
+from ..testing.util import datetime_now_utc_naive, date_now_utc_naive
```
matriv, line 58: why this is relative?

src/crate/testing/util.py
```diff
@@ -0,0 +1,30 @@
...
+# under the License.
+#
+# However, if you have executed another commercial license agreement
+# with Crate these terms will supersede the license and you may use the
+# software solely pursuant to the terms of the relevant commercial agreement.
+
+from datetime import datetime, timezone
+
+
+def datetime_now_utc_naive():
```
matriv, line 25: why did you choose `naive` in the naming?

matriv (commented):
Thx @amotl left a couple of comments.

src/crate/client/tests.py
amotl: PyCharm created that import line for me. The import statements above are also relative.

src/crate/testing/util.py
amotl: The documentation where this patch is based upon also uses the same nomenclature "aware" vs. "naive".

- https://docs.python.org/3/library/datetime.html
- https://blog.ganssle.io/articles/2019/11/utcnow.html

src/crate/client/sqlalchemy/doctests/itests.txt
```diff
@@ -65,10 +65,10 @@ Retrieve the location from the database::
     >>> location.name
     'Earth'
 
-Date should have been set at the insert due to default value via python method::
+Date should have been set at the insert due to default value via Python method::
 
-    >>> from datetime import datetime
-    >>> now = datetime.utcnow()
+    >>> from datetime import datetime, timezone
+    >>> now = datetime.now(timezone.utc).replace(tzinfo=None)
```
mfussenegger, line 71: I don't understand the change here.  Isn't `now(timezone.utc).replace(tzinfo=None)` the same as `utcnow()`?

src/crate/client/sqlalchemy/dialect.py
```diff
@@ -90,7 +90,8 @@ def process(value):
             if not value:
                 return
             try:
-                return datetime.utcfromtimestamp(value / 1e3).date()
+                return datetime.fromtimestamp(value / 1e3, tz=timezone.utc) \
+                    .replace(tzinfo=None).date()
```
mfussenegger, line 94: If we throw away the tzinfo why not keep using utcfromtimestamp?

src/crate/testing/util.py
```diff
@@ -0,0 +1,30 @@
...
+    return datetime.now(timezone.utc).replace(tzinfo=None)
```
mfussenegger, line 26: Isn't this just utcnow?

mfussenegger (changes requested):

src/crate/client/sqlalchemy/dialect.py
matriv: I think the idea is to be a preparation step to actually allow the user to pass a tzinfo here, later on?

src/crate/client/sqlalchemy/doctests/itests.txt
amotl: It is really sad. In normal situations, or when the system time zone setting is expressed in UTC, everything is fine. However, it is not the case if the system time zone is anything other than UTC.

I've tried to improve the situation to outline the problem with c1f183ed4.

mfussenegger: This is not the case.

[utcnow](https://docs.python.org/3/library/datetime.html?highlight=datetime#datetime.datetime.utcnow):

> Return the current UTC date and time, with [tzinfo](https://docs.python.org/3/library/datetime.html?highlight=datetime#datetime.datetime.tzinfo) None.

Yes, it is a naive datetime without timezone info attached, but it _is_ using utc.

```
>>> print(datetime.utcnow(), '\n', datetime.now(tz=timezone.utc).replace(tzinfo=None), '\n', datetime.now())
2022-07-22 18:27:27.055192
 2022-07-22 18:27:27.055200
 2022-07-22 20:27:27.055213
```


```
↪  date +"%Z %z"
CEST +0200
```

mfussenegger: Also as the articles point out, the problem is the use of the naive datetime (lack of tzinfo) in subsequent operations. If we remove the tzinfo anyways via `.replace(tzinfo=None)` we don't gain anything by using `now(tz=timezone.utc)` but only make things more complicated.

amotl: Thank you. You convinced me that this goes too far. I've submitted #441 instead, which solely focuses on mitigating a fluke I ~~regularly~~ occasionally observed when running the test suite on CI after midnight.

PR #441, opened 2022-07-22
Tests: Use UTC-based `utcnow().date()` instead of localtime-based `date.today()`

amotl:
### Motivation
- Hunting an occasional known fluke when running the test suite on CI after midnight.
- Further set the stage for #437.
- Back out of #440, which went too far.

### About

`datetime.date.today()` returns the current local date. So, its outcome depends on the system time zone. Because we want to use UTC-based date and time objects instead, `datetime.utcnow().date()` is a better choice.

### DIFF
--- a/src/crate/client/sqlalchemy/doctests/itests.txt
+++ b/src/crate/client/sqlalchemy/doctests/itests.txt
@@ -105,7 +105,7 @@ aren't set when the row is inserted as there is no default method::
 
 The datetime and date can be set using a update statement::
 
-    >>> location.nullable_date = datetime.today()
+    >>> location.nullable_date = datetime.utcnow().date()
     >>> location.nullable_datetime = datetime.utcnow()
     >>> session.flush()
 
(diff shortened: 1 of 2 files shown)


src/crate/testing/test_datetime.py
```diff
@@ -0,0 +1,90 @@
...
+import os
+import sys
+import unittest
+from datetime import datetime, date
+from unittest import TestCase, mock
+
+import time_machine
+
+
+class DateTodayTest(TestCase):
```
mfussenegger, line 31: Please remove, there's no need to create tests that verify documented behavior of the standard library.

mfussenegger (commented):
Other than the test addition this looks good to me.

The commit message could be clarified a bit. Using UTC is not better - it depends on the use case.

What is a problem is that we were comparing utc variants with local variants in https://github.com/crate/crate-python/blob/a849ceaaf6855df7d347197ff1d7edf062a9db54/src/crate/client/sqlalchemy/doctests/itests.txt#L70-L84

That is the problem and caused test flakyness.

src/crate/testing/test_datetime.py
amotl: Thanks. I've removed the test cases and amended the commit message.

mfussenegger (approved):

PR #442, opened 2022-08-05
Add a generic data type converter to the `Cursor` object

amotl:
Hi there,

### Introduction
After converging #395 to #437, it became clear / emerged that we wanted to have the type conversion

- only being optionally enabled with a feature flag /cc @mfussenegger 
- being implemented more elegantly /cc @matriv 

This patch aims to resolve both aspects.

### Details
This will allow converting fetched data from CrateDB data types to Python data types in different ways. Its usage is completely optional. When not used, the feature will not incur any overhead.

There is also a `DefaultTypeConverter`, which aims to become a sane default choice when looking at this kind of convenience. It will enable the user to work with native Python data types from the start, for all database types where this makes sense, without needing to establish the required set of default value converters on their own.

If the `DefaultTypeConverter` does not fit the user's aims, it is easy to define custom type converters, possibly reusing specific ones from the library.

### Acknowledgements
The patch follows the implementation suggested by the [PyAthena](https://github.com/laughingman7743/PyAthena) driver very closely, see [PyAthena/converter.py](https://github.com/laughingman7743/PyAthena/blob/master/pyathena/converter.py). Thank you!

### Synopsis
```python
>>> cursor = connection.cursor(converter=Cursor.get_default_converter())
>>> cursor.execute(stmt)
>>> cursor.fetchone()
['foo', IPv4Address('10.10.10.1'), datetime.datetime(2022, 7, 18, 18, 10, 36, 758000)]
```

### Thoughts
In the form of an early review, please let me know if you endorse this approach and which adjustments you would like to see. 

Personally, I would add more test cases to `test_cursor.py`, in order to exercise the machinery in more detail and to increase code coverage, and probably make the functionality more present in the documentation.

Also, as @mfussenegger mentioned at https://github.com/crate/crate-python/pull/437/files#r930822457, it might make sense to also add the important aspect of converting elements within `ARRAY` types to this patch already.

With kind regards,
Andreas.

---

### Backlog
- [x] Make the functionality more present in the documentation. See 1acc44e1.
- [x] Setting the timezone for datetime objects. See #445. /cc @mfussenegger
- [ ] Decide if the type converter for the `BIT` data type should be included into the `DefaultTypeConverter`.
  https://github.com/crate/crate-python/blob/33ecbf36a3b50606f5138ce9dcee6b00b81996c0/src/crate/client/doctests/cursor.txt#L347

### DIFF
--- a/src/crate/client/connection.py
+++ b/src/crate/client/connection.py
@@ -46,6 +46,7 @@ class Connection(object):
                  socket_tcp_keepidle=None,
                  socket_tcp_keepintvl=None,
                  socket_tcp_keepcnt=None,
+                 converter=None,
                  ):
         """
         :param servers:
@@ -99,7 +100,13 @@ class Connection(object):
             Set the ``TCP_KEEPCNT`` socket option, which overrides
             ``net.ipv4.tcp_keepalive_probes`` kernel setting if ``socket_keepalive``
             is ``True``.
+        :param converter:
+            (optional, defaults to ``None``)
+            A `Converter` object to propagate to newly created `Cursor` objects.
         """
+
+        self._converter = converter
+
         if client:
             self.client = client
         else:
@@ -123,12 +130,16 @@ class Connection(object):
         self.lowest_server_version = self._lowest_server_version()
         self._closed = False
 
-    def cursor(self):
+    def cursor(self, **kwargs) -> Cursor:
         """
         Return a new Cursor Object using the connection.
         """
+        converter = kwargs.pop("converter", self._converter)
         if not self._closed:
-            return Cursor(self)
+            return Cursor(
+                connection=self,
+                converter=converter,
+            )
         else:
             raise ProgrammingError("Connection closed")
 
--- a/src/crate/client/converter.py
+++ b/src/crate/client/converter.py
@@ -1,0 +1,135 @@
+# -*- coding: utf-8; -*-
+#
+# Licensed to CRATE Technology GmbH ("Crate") under one or more contributor
+# license agreements.  See the NOTICE file distributed with this work for
+# additional information regarding copyright ownership.  Crate licenses
+# this file to you under the Apache License, Version 2.0 (the "License");
+# you may not use this file except in compliance with the License.  You may
+# obtain a copy of the License at
+#
+#   http://www.apache.org/licenses/LICENSE-2.0
+#
+# Unless required by applicable law or agreed to in writing, software
+# distributed under the License is distributed on an "AS IS" BASIS, WITHOUT
+# WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.  See the
+# License for the specific language governing permissions and limitations
+# under the License.
+#
+# However, if you have executed another commercial license agreement
+# with Crate these terms will supersede the license and you may use the
+# software solely pursuant to the terms of the relevant commercial agreement.
+"""
+Machinery for converting CrateDB database types to native Python data types.
+
+https://crate.io/docs/crate/reference/en/latest/interfaces/http.html#column-types
+"""
+import ipaddress
+from copy import deepcopy
+from datetime import datetime
+from enum import Enum
+from typing import Any, Callable, Dict, List, Optional, Union
+
+ConverterFunction = Callable[[Optional[Any]], Optional[Any]]
+ColTypesDefinition = Union[int, List[Union[int, "ColTypesDefinition"]]]
+
+
+def _to_ipaddress(value: Optional[str]) -> Optional[Union[ipaddress.IPv4Address, ipaddress.IPv6Address]]:
+    """
+    https://docs.python.org/3/library/ipaddress.html
+    """
+    if value is None:
+        return None
+    return ipaddress.ip_address(value)
+
+
+def _to_datetime(value: Optional[float]) -> Optional[datetime]:
+    """
+    https://docs.python.org/3/library/datetime.html
+    """
+    if value is None:
+        return None
+    return datetime.utcfromtimestamp(value / 1e3)
+
+
+def _to_default(value: Optional[Any]) -> Optional[Any]:
+    return value
+
+
+# Symbolic aliases for the numeric data type identifiers defined by the CrateDB HTTP interface.
+# https://crate.io/docs/crate/reference/en/latest/interfaces/http.html#column-types
+class DataType(Enum):
+    NULL = 0
+    NOT_SUPPORTED = 1
+    CHAR = 2
+    BOOLEAN = 3
+    TEXT = 4
+    IP = 5
+    DOUBLE = 6
+    REAL = 7
+    SMALLINT = 8
+    INTEGER = 9
+    BIGINT = 10
+    TIMESTAMP_WITH_TZ = 11
+    OBJECT = 12
+    GEOPOINT = 13
+    GEOSHAPE = 14
+    TIMESTAMP_WITHOUT_TZ = 15
+    UNCHECKED_OBJECT = 16
+    REGPROC = 19
+    TIME = 20
+    OIDVECTOR = 21
+    NUMERIC = 22
+    REGCLASS = 23
+    DATE = 24
+    BIT = 25
+    JSON = 26
+    CHARACTER = 27
+    ARRAY = 100
+
+
+ConverterMapping = Dict[DataType, ConverterFunction]
+
+
+# Map data type identifier to converter function.
+_DEFAULT_CONVERTERS: ConverterMapping = {
+    DataType.IP: _to_ipaddress,
+    DataType.TIMESTAMP_WITH_TZ: _to_datetime,
+    DataType.TIMESTAMP_WITHOUT_TZ: _to_datetime,
+}
+
+
+class Converter:
+    def __init__(
+        self,
+        mappings: Optional[ConverterMapping] = None,
+        default: ConverterFunction = _to_default,
+    ) -> None:
+        self._mappings = mappings or {}
+        self._default = default
+
+    def get(self, type_: ColTypesDefinition) -> ConverterFunction:
+        if isinstance(type_, int):
+            return self._mappings.get(DataType(type_), self._default)
+        type_, inner_type = type_
+        if DataType(type_) is not DataType.ARRAY:
+            raise ValueError(f"Data type {type_} is not implemented as collection type")
+
+        inner_convert = self.get(inner_type)
+
+        def convert(value: Any) -> Optional[List[Any]]:
+            if value is None:
+                return None
+            return [inner_convert(x) for x in value]
+
+        return convert
+
+
+class DefaultTypeConverter(Converter):
+    def __init__(self, more_mappings: Optional[ConverterMapping] = None) -> None:
+        mappings: ConverterMapping = {}
+        mappings.update(deepcopy(_DEFAULT_CONVERTERS))
+        if more_mappings:
+            mappings.update(deepcopy(more_mappings))
+        super().__init__(
+            mappings=mappings, default=_to_default
+        )
--- a/src/crate/client/cursor.py
+++ b/src/crate/client/cursor.py
@@ -19,8 +19,10 @@
 # with Crate these terms will supersede the license and you may use the
 # software solely pursuant to the terms of the relevant commercial agreement.
 
-from .exceptions import ProgrammingError
 import warnings
+
+from .converter import Converter
+from .exceptions import ProgrammingError
 
 
 class Cursor(object):
@@ -30,9 +32,10 @@ class Cursor(object):
     """
     lastrowid = None  # currently not supported
 
-    def __init__(self, connection):
+    def __init__(self, connection, converter: Converter):
         self.arraysize = 1
         self.connection = connection
+        self._converter = converter
         self._closed = False
         self._result = None
         self.rows = None
@@ -50,7 +53,10 @@ class Cursor(object):
         self._result = self.connection.client.sql(sql, parameters,
                                                   bulk_parameters)
         if "rows" in self._result:
-            self.rows = iter(self._result["rows"])
+            if self._converter is None:
+                self.rows = iter(self._result["rows"])
+            else:
+                self.rows = iter(self._convert_rows())
 
     def executemany(self, sql, seq_of_parameters):
         """
@@ -73,9 +79,13 @@ class Cursor(object):
             "duration": sum(durations) if durations else -1,
             "rows": [],
             "cols": self._result.get("cols", []),
+            "col_types": self._result.get("col_types", []),
             "results": self._result.get("results")
         }
-        self.rows = iter(self._result["rows"])
+        if self._converter is None:
+            self.rows = iter(self._result["rows"])
+        else:
+            self.rows = iter(self._convert_rows())
         return self._result["results"]
 
     def fetchone(self):
@@ -210,3 +220,24 @@ class Cursor(object):
                 "duration" not in self._result:
             return -1
         return self._result.get("duration", 0)
+
+    def _convert_rows(self):
+        """
+        Iterate rows, apply type converters, and generate converted rows.
+        """
+        assert "col_types" in self._result and self._result["col_types"], \
+               "Unable to apply type conversion without `col_types` information"
+
+        # Resolve `col_types` definition to converter functions. Running the lookup
+        # redundantly on each row loop iteration would be a huge performance hog.
+        types = self._result["col_types"]
+        converters = [
+            self._converter.get(type) for type in types
+        ]
+
+        # Process result rows with conversion.
+        for row in self._result["rows"]:
+            yield [
+                convert(value)
+                for convert, value in zip(converters, row)
+            ]
--- a/src/crate/client/doctests/cursor.txt
+++ b/src/crate/client/doctests/cursor.txt
@@ -2,9 +2,18 @@
 Cursor
 ======
 
+
+Setup
+=====
+
+This section sets up a cursor object, inspects some of its attributes, and sets
+up the response for subsequent cursor operations.
+
 ::
 
     >>> from crate.client import connect
+    >>> from crate.client.converter import DefaultTypeConverter
+    >>> from crate.client.cursor import Cursor
 
     >>> connection = connect(client=connection_client_mocked)
     >>> cursor = connection.cursor()
@@ -81,7 +90,7 @@ If the specified number of rows not being available, fewer rows may returned::
 
     >>> cursor.execute('')
 
-If no number of rows are specified it defaults to the current cursor.arraysize::
+If no number of rows are specified it defaults to the current ``cursor.arraysize``::
 
     >>> cursor.arraysize
     1
@@ -179,7 +188,7 @@ The attribute is -1 in case the cursor has been closed::
     >>> cursor.rowcount
     -1
 
-If the last respsonse doesn't contain the rowcount attribute -1 is returned::
+If the last response does not contain the rowcount attribute, ``-1`` is returned::
 
     >>> cursor = connection.cursor()
     >>> connection.client.set_next_response({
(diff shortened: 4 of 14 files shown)


src/crate/client/cursor.py
```diff
@@ -210,3 +218,23 @@ def duration(self):
...
+        Iterate rows, apply type converters, and generate converted rows.
+        """
+        assert "col_types" in self._result, \
+            "Unable to apply type conversion without `col_types` information"
+        type_id_list = self._result["col_types"]
+        for row in self._result["rows"]:
+            yield [
+                self._converter.convert(type_id, value)
+                for type_id, value in zip(type_id_list, row)
+            ]
```
amotl, line 233: Hi @matriv,

at https://github.com/crate/crate-python/pull/437#discussion_r924614148, you asked:

> Can we do [this] more elegant?

Maybe this implementation fits your eyes better.

With kind regards,
Andreas.

src/crate/client/doctests/cursor.txt
```diff
@@ -296,6 +304,69 @@ closed connection an ``ProgrammingError`` exception will be raised::
...
+
+CrateDB's ``BIT`` data type has the numeric identifier ``25``. To create a
+simple converter for converging CrateDB's ``BIT`` type to Python's ``int``
+type::
+
+    >>> from crate.client.converter import Converter
+
+    >>> converter = Converter()
+    >>> converter.set(25, lambda value: int(value[2:-1], 2))
+    >>> cursor = connection.cursor(converter=converter)
```
amotl, line 348: @matriv: Simple converters can easily be implemented with lambdas.

seut:
👍 I really like this approach and would definitely prefer it over #437.

src/crate/client/converter.py
```diff
@@ -0,0 +1,96 @@
...
+def _to_ipaddress(value: Optional[str]) -> Optional[Union[ipaddress.IPv4Address, ipaddress.IPv6Address]]:
+    """
+    https://docs.python.org/3/library/ipaddress.html
+    """
+    if value is None:
+        return None
+    return ipaddress.ip_address(value)
+
+
+def _to_datetime(value: Optional[str]) -> Optional[datetime]:
```
mfussenegger, line 41: type annotation is wrong here. A `str` would result in an error with `utcfromtimestamp(value / 1e3)`

src/crate/client/converter.py
```diff
@@ -0,0 +1,96 @@
...
+    return datetime.utcfromtimestamp(value / 1e3)
+
+
+def _to_default(value: Optional[str]) -> Optional[str]:
+    return value
+
+
+# Mapping for numeric data type identifiers defined by the CrateDB HTTP interface.
+# https://crate.io/docs/crate/reference/en/latest/interfaces/http.html#column-types
+_DEFAULT_CONVERTERS: Dict[int, Callable[[Optional[int]], Optional[Any]]] = {
```
mfussenegger, line 56: This type annotation results in mypy errors because `_to_ipaddress` and `_to_datetime` have different signatures

src/crate/client/converter.py
```diff
@@ -0,0 +1,96 @@
...
+        return self.mappings.get(type_, self._default)
+
+    def set(self, type_: int, converter: Callable[[Optional[int]], Optional[Any]]) -> None:
+        self.mappings[type_] = converter
+
+    def remove(self, type_: int) -> None:
+        self.mappings.pop(type_, None)
+
+    def update(self, mappings: Dict[int, Callable[[Optional[int]], Optional[Any]]]) -> None:
+        self.mappings.update(mappings)
```
mfussenegger, line 85: Not sure if we want to make this mutable?
I also wonder if we should introduce an `Enum` or similar to give names to the type ids?

src/crate/client/doctests/cursor.txt
```diff
@@ -296,6 +304,69 @@ closed connection an ``ProgrammingError`` exception will be raised::
...
+
+    >>> connection.client.set_next_response({
+    ...     "col_types": [4, 5, 11],
+    ...     "rows":[ [ "foo", "10.10.10.1", 1658167836758 ] ],
+    ...     "cols":[ "name", "address", "timestamp" ],
+    ...     "rowcount":1,
+    ...     "duration":123
+    ... })
+
+    >>> cursor.execute('')
```
mfussenegger, line 327: I think this would confuse users because this is a mocked cursor and `client.set_next_response` isn't something that they can use.

Either we need to point this out explicitly, hide it or use real data and don't mock it.

mfussenegger (commented):
Left a couple of comments regarding some of the implementation details.


I'm also not sure about the level of separation. E.g. the converter application could probably be separated from the cursor without much loss of API/UI convenience. But no strong opinion here. 

On the other hand, I think something common like setting the timezone for datetime objects could be made even easier for users.

mfussenegger:
There is btw. also https://peps.python.org/pep-0249/#type-objects in the DB API definition - I didn't have a closer look at it, but we should double check to see if it affects what the PR would be doing.

amotl:
Thank you for endorsing this patch and for your valuable suggestions.

src/crate/client/cursor.py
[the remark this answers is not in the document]
amotl: Hi. It was just by accident. I just made the method private.

src/crate/client/converter.py
amotl: Oh, right. Thank for spotting it. After making some adjustments in this area, `mypy src/crate/client/converter.py` doesn't complain anymore now.

src/crate/client/converter.py
amotl: Fixed. Thank you!

src/crate/client/converter.py
amotl: > Not sure if we want to make this mutable?

I also wasn't sure, but liked the flexibility in the end. See also my other comment below.

> I also wonder if we should introduce an `Enum` or similar to give names to the type ids?

Good idea. I've introduced an `Enum` with 00dbb42380.

src/crate/client/doctests/cursor.txt
amotl: At the beginning of the file `cursor.txt`, there is a corresponding note about this detail.

> Hardcode the next response of the mocked connection client, so we won't need a sql statement
to execute.

After that, the `set_next_response()` method is already used five times within `cursor.txt`, so I didn't hesitate to continue implementing the test cases this way.

amotl:
Hi again,

other than the inline comments, let me address further questions here.

> Personally, I would add more test cases to `test_cursor.py`, in order to exercise the machinery in more detail and to increase code coverage.

Done, mostly with 141054780f.

> It might make sense to also add the important aspect of converting elements within ARRAY types to this patch already.

Done with b21c38e99.

> I'm also not sure about the level of separation. E.g. the converter application could probably be separated from the cursor without much loss of API/UI convenience. But no strong opinion here.

I also thought back and forth about this detail. In the end, I followed the path outlined by https://github.com/laughingman7743/PyAthena very closely, where I originally discovered this way of doing it. See [PyAthena/converter.py](https://github.com/laughingman7743/PyAthena/blob/master/pyathena/converter.py).

> On the other hand, I think something common like setting the timezone for datetime objects could be made even easier for users.

Indeed, that would be sweet. I will think about how we could add this feature.

> There is btw. also https://peps.python.org/pep-0249/#type-objects in the DB API definition - I didn't have a closer look at it, but we should double check to see if it affects what the PR would be doing.

I also found that section in the DBAPI spec, but I think it is exclusively about **input** parameter conversion:

> Many databases need to have the input in a particular format for binding to an operation’s input parameters. For example, if an input is destined for a DATE column, then it must be bound to the database in a particular string format. 

On the other hand, the current implementation is exclusively about **output** data conversion. That being said, the converter functions itself can well be reused for that use case, but otherwise I think it is a different area, and to be addressed within a different patch.

With kind regards,
Andreas.

---

### Backlog
- Make the functionality more present in the documentation.
- Setting the timezone for datetime objects. /cc @mfussenegger 
- Think about an option to enable the `DefaultTypeConverter` using an SQLAlchemy DBURI connection parameter, because in such scenarios, the implementer would not be able to touch any code at all. This is in order to close the loop where @Aymaru was probably coming from, see #395 and #437.

src/crate/client/converter.py
```diff
@@ -0,0 +1,147 @@
...
+    return ipaddress.ip_address(value)
+
+
+def _to_datetime(value: Optional[float]) -> Optional[datetime]:
+    """
+    https://docs.python.org/3/library/datetime.html
+    """
+    if value is None:
+        return None
+    return datetime.utcfromtimestamp(value / 1e3)
```
amotl, line 50: Dear @mfussenegger, @matriv, and @seut,

regarding @mfussenegger's proposal to let the user optionally specify a time zone, and make the machinery return a timezone-aware datetime object, I see two options here.

Let's assume the user defines a custom timezone object like
```python
>>> import datetime
>>> tz_mst = datetime.timezone(datetime.timedelta(hours=7), name="MST")
```

and a timestamp in Epoch like
```python
# Fri, 16 Sep 2022 11:09:16 GMT
timestamp = 1663326556
```

On this spot, which variant would be semantically the right one? Note that [`datetime.utcfromtimestamp`](https://docs.python.org/3/library/datetime.html#datetime.datetime.utcfromtimestamp) does not accept the `tz` keyword argument, it will always return a naive datetime object with `tzinfo=None`.

1. We can either use [`datetime.fromtimestamp`](https://docs.python.org/3/library/datetime.html#datetime.datetime.fromtimestamp), which **does** accept the `tz` keyword argument.
   ```python
   >>> datetime.datetime.fromtimestamp(timestamp, tz=tz_mst)
   datetime.datetime(2022, 9, 16, 18, 9, 16, tzinfo=datetime.timezone(datetime.timedelta(seconds=25200), 'MST'))
   ```

2. Or we can replace the `tzinfo` on the naive datetime object returned by `utcfromtimestamp`, making it timezone-aware.
   ```python
   >>> datetime.datetime.utcfromtimestamp(timestamp).replace(tzinfo=tz_mst)
   datetime.datetime(2022, 9, 16, 11, 9, 16, tzinfo=datetime.timezone(datetime.timedelta(seconds=25200), 'MST'))
   ```

I think option 1. is the right choice. So the implementation would look like something along the lines of:
```python

def set_timezone(self, tz):
  if not isinstance(tz, datetime.timezone):
    raise TypeError(f"Timezone object has wrong type: {tz}")
  self.timezone = tz

def _to_datetime(self, value):
  """
  Convert CrateDB's `TIMESTAMP` column to the native Python `datetime` object.
  When a timezone is given, return a timezone-aware object.
  Otherwise, return a naive object which is meant to be the time in the UTC time zone.
  """
  if self.timezone is not None:
    return datetime.fromtimestamp(value / 1e3, tz=self.timezone)
  else:
    return datetime.utcfromtimestamp(value / 1e3)
```

What do you think?

With kind regards,
Andreas.

amotl: Another implementation could look like this, it is **always** using `datetime.fromtimestamp(value / 1e3, tz=self.timezone)`, where `self.timezone` would be `datetime.timezone.utc` by default, i.e. it will always return timezone-aware datetime objects.

While definitively easier and more straight-forward, it could be less performant when **not** aiming to use timezone-aware datetime objects and stay in "naive" land **on purpose**.

```python
def __init__(self):
  # Return datetype objects in UTC by default.
  self.timezone = datetime.timezone.utc

def set_timezone(self, tz):
  """
  Specify a time zone if you want to have your datetime objects in a time zone other than UTC.
  """
  if not isinstance(tz, datetime.timezone):
    raise TypeError(f"Timezone object has wrong type: {tz}")
  self.timezone = tz

def _to_datetime(self, value):
  """
  Convert CrateDB's `TIMESTAMP` column to a timezone-aware native Python `datetime` object.
  When no timezone is given, UTC is used.
  """
  return datetime.fromtimestamp(value / 1e3, tz=self.timezone)
```

amotl: > definitively easier and more straight-forward

When thinking about it once more, I think I would prefer the implementation variant outlined in my latest post. In this way, it is always consistent that timezone-aware datetime objects are returned from the data converter machinery.

> it could be less performant when not aiming to use timezone-aware datetime objects and stay in "naive" land on purpose.

I think when users are aiming for raw speed, they would completely opt-out of the data type converter machinery at all.

amotl: Hi again. Talk is cheap. I've converged a few more thoughts about this into a proposed patch at #445, completely separating the topic of timezone-aware `datetime` handling from this one.

src/crate/client/converter.py
```diff
@@ -0,0 +1,147 @@
...
+
+
+class Converter:
+    def __init__(
+        self,
+        mappings: Dict[int, Callable[[Optional[InputVal]], Optional[Any]]] = None,
+        default: Callable[[Optional[InputVal]], Optional[Any]] = _to_default,
+    ) -> None:
+        self._mappings = mappings or {}
+        self._default = default
```
mfussenegger, line 102: Could promote usage of the enum further and use it in `mappings` as key instead of the `int` value.

That would implicitly document what `int` values are acceptable.

src/crate/client/converter.py
mfussenegger: > I also wasn't sure, but liked the flexibility in the end to make the user able to add and remove converters as they like.

I'd keep it immutable for now. It reduces the API surface and makes it easier to reason about the mappings.
And _if_ a legitimate use-case pops up, it's easy to change later. On the other hand, changing a mutable mapping to immutable later on would be a breaking change.

src/crate/client/doctests/cursor.txt
mfussenegger: Is there a reason we can't/shouldn't hide it for the rendered docs?

src/crate/client/converter.py
amotl: Are you fine with removing mutability by 299b7713ad?

src/crate/client/doctests/cursor.txt
amotl: As far as I can see, `cursor.txt` is not part of the rendered docs at https://crate.io/docs/python/, it is only used as a doctest.

src/crate/client/converter.py
```diff
@@ -0,0 +1,147 @@
...
+    return datetime.utcfromtimestamp(value / 1e3)
+
+
+def _to_default(value: Optional[InputVal]) -> Optional[Any]:
+    return value
+
+
+# Symbolic aliases for the numeric data type identifiers defined by the CrateDB HTTP interface.
+# https://crate.io/docs/crate/reference/en/latest/interfaces/http.html#column-types
+class CrateDatatypeIdentifier(Enum):
```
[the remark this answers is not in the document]
amotl: Thanks. Renamed with 134d634.

src/crate/client/converter.py
amotl: Thank you for bringing this up.

### Problem
Without looking further into the details and the state of the onion, I was wary of using an Enum-type as key, because I found different resources like those referenced below, which indicate possible performance losses. Especially because the converter sits in a busy place of the driver, I decided to not use it this way and use an `int` as key instead.

### Thoughts
If we still want to use an Enum-type as a key on the mapping data structure itself, I can think of two options, but I will be happy to hear if you see others.
- Use in-memory `@memoize`-style caching on the lookup path, if that would work around the Enum-lookup-is-slow problem.
- Look into using the `fastenum` library.

### References

TLDR; Apparently, Python 3.4's first Enum implementation was pretty slow. While it got better with subsequent releases, the `fastenum` library still claims to be up to 3.5x faster in benchmarks on recent versions of Python.

- https://quantlane.com/blog/fastenum/, https://gitlab.com/quantlane/libs/fastenum
- https://pypi.org/project/fast-enum/
- https://bugs.python.org/issue39102

mfussenegger: Is this relevant for how we use it?

What we should do independent of this is to  move the dict lookup out of the row loop. The columns/types are the same across all rows. We can retrieve the converter for all columns once and then process all rows with them.

Then you've 1 enum attribute access and 1 dict lookup *per* column. That shouldn't be a performance issue.

src/crate/client/tests.py
```diff
@@ -335,6 +312,10 @@ def _try_execute(cursor, stmt):
     try:
         cursor.execute(stmt)
     except Exception:
+        # FIXME: Why does this croak on statements like ``DROP TABLE cities``?
+        # Note: When needing to debug the test environment, you may want to
+        #       enable this logger statement.
+        # log.exception("Executing SQL statement failed")
```
amotl, line 318: ### About
Tests: Add inline comment about how to debug the test environment.

### Background story
It was needed because the tests suddenly started failing mysteriously on my machine.

Out of the blue, the test suite started croaking like `TABLE 'locations' already exists`, only to find that CrateDB would not accept any write operations after finding that my disk usage was 96%, while I had still 60G of free disk space.

Unfortunately, I did not discover this easily, because the test suite swallowed the exception on this spot. When enabling the `log.exception` statement, the error response from CrateDB became immediately visible.

src/crate/client/converter.py
```diff
@@ -0,0 +1,159 @@
...
+    CHARACTER = 27
+    ARRAY = 100
+
+
+# Map data type identifier to converter function.
+_DEFAULT_CONVERTERS: Dict[DataType, ConverterFunction] = {
+    DataType.IP: _to_ipaddress,
+    DataType.TIMESTAMP_WITH_TZ: _to_datetime,
+    DataType.TIMESTAMP_WITHOUT_TZ: _to_datetime,
+}
```
amotl, line 97: Not sure if the `DefaultTypeConverter` should apply the `_to_datetime` function to _both_ kinds of `TIMESTAMP` types?

src/crate/client/doctests/cursor.txt
```diff
@@ -296,6 +304,69 @@ closed connection an ``ProgrammingError`` exception will be raised::
...
+    ...     "rows":[ [ "B'0110'" ] ],
+    ...     "cols":[ "value" ],
+    ...     "rowcount":1,
+    ...     "duration":123
+    ... })
+
+    >>> cursor.execute('')
+
+    >>> cursor.fetchone()
+    [6]
```
amotl, line 364: Do you think the type converter for the `BIT` data type should be included into the `DefaultTypeConverter`, like the converters for `IP` and `TIMESTAMP`? I am talking about this one:

https://github.com/crate/crate-python/blob/33ecbf36a3b50606f5138ce9dcee6b00b81996c0/src/crate/client/doctests/cursor.txt#L347

src/crate/client/doctests/cursor.txt
amotl: I've created #448 to have a separate discussion about the topic how to improve/adjust doctests vs. rendered documentation.

src/crate/client/converter.py
```diff
@@ -0,0 +1,159 @@
...
+        self,
+        mappings: Optional[Dict[DataType, ConverterFunction]] = None,
+        default: ConverterFunction = _to_default,
+    ) -> None:
+        self._mappings = mappings or {}
+        self._default = default
+
+    @property
+    def mappings(self) -> Dict[DataType, ConverterFunction]:
+        return self._mappings
```
mfussenegger, line 111:
```suggestion
```

src/crate/client/converter.py
```diff
@@ -0,0 +1,159 @@
...
+import ipaddress
+from copy import deepcopy
+from datetime import datetime
+from enum import Enum
+from typing import Any, Callable, Dict, List, Optional, Tuple, Union
+
+InputVal = Any
+ConverterFunction = Callable[[Optional[InputVal]], Optional[Any]]
+ConverterDefinition = Union[ConverterFunction, Tuple[ConverterFunction, "ConverterDefinition"]]
+ColTypesDefinition = Union[int, List[Union[int, "ColTypesDefinition"]]]
```
mfussenegger, line 35:
```suggestion
ConverterFunction = Callable[[Optional[Any]], Optional[Any]]
ConverterDefinition = Union[ConverterFunction, Tuple[ConverterFunction, "ConverterDefinition"]]
ColTypesDefinition = Union[int, List[Union[int, "ColTypesDefinition"]]]
```

No need for custom vocabulary for `Any`?

src/crate/client/doctests/cursor.txt
amotl: I totally agree with you. I think your guidance should be implemented seperately, being tracked by #448. If you want to see it sooner than later, than I will be happy to tackle it before coming back to this patch.

mfussenegger: Sounds good to handle this separately in a follow up

src/crate/client/converter.py
```diff
@@ -0,0 +1,159 @@
...
+
+    @property
+    def mappings(self) -> Dict[DataType, ConverterFunction]:
+        return self._mappings
+
+    def get(self, type_: DataType) -> ConverterFunction:
+        return self.mappings.get(type_, self._default)
+
+    def set(self, type_: DataType, converter: ConverterFunction) -> None:
+        self.mappings[type_] = converter
```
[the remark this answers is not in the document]
amotl: I've just added 14a355daa, to follow your suggestion.

src/crate/client/doctests/cursor.txt
mfussenegger: I'd leave it out for now. Easy enough to add later if requested by users

src/crate/client/connection.py
```diff
@@ -123,12 +130,16 @@ def __init__(self,
         self.lowest_server_version = self._lowest_server_version()
         self._closed = False
 
-    def cursor(self):
+    def cursor(self, cursor=None, **kwargs) -> Cursor:
```
mfussenegger, line 133: What's the `cursor=None` for?

src/crate/client/cursor.py
```diff
@@ -210,3 +221,31 @@ def duration(self):
...
+                convert(value)
+                for convert, value in zip(converters, row)
+            ]
+
+    @staticmethod
+    def get_default_converter(mappings: Optional[ConverterMapping] = None) -> Converter:
+        """
+        Return the standard converter instance.
+        """
+        return DefaultTypeConverter(more_mappings=mappings)
```
mfussenegger, line 251: Is this needed? It's just wrapping a constructor which could be called directly?

mfussenegger (commented):
Left two more comments, otherwise looks good to me

src/crate/client/connection.py
amotl: It probably slipped in from the role model implementation of PyAthena. I think it can be removed. Thanks for spotting it.

```
def cursor(self, cursor: Optional[Type[BaseCursor]] = None, **kwargs) -> BaseCursor:
```

-- https://github.com/laughingman7743/PyAthena/blob/2f2c0e11487426bd382f480a8468e9bc78c42156/pyathena/connection.py#L231

src/crate/client/cursor.py
amotl: It looks like `get_default_converter()` is currently never called with any parameters, so I guess it can also be removed. Thanks!

mfussenegger (approved):

PR #445, opened 2022-09-16
Return `TIMESTAMP` types as timezone-aware native Python `datetime` objects

amotl:
### About
When requested, make the driver return timezone-aware native Python `datetime` objects for CrateDB's `TIMESTAMP` types. The user can easily supply the desired time zone by different means.

### Rationale
@mfussenegger asked at https://github.com/crate/crate-python/pull/442#pullrequestreview-1088223785:
> I think something common like setting the timezone for datetime objects could be made even easier for users.

I shared some initial thoughts about it at https://github.com/crate/crate-python/pull/442#pullrequestreview-1110538218 ff., and finally settled with this implementation, separately from the baseline implementation for the data type converter.

### Details
This patch adds additional convenience based on the data type converter machinery added with #442.

- A `time_zone` _keyword argument_ can be passed to both the `connect()` method, or when creating new `Cursor` objects.
- The `time_zone` _attribute_ can also be changed at runtime on both the `connection` and `cursor` object instances.

Different ways to populate the `time_zone` value are supported. Some examples::

- ``datetime.timezone.utc``
- ``datetime.timezone(datetime.timedelta(hours=7), name="MST")``
- ``pytz.timezone("Australia/Sydney")``
- ``zoneinfo.ZoneInfo("Australia/Sydney")``
- ``+0530`` (UTC offset in string format)

### Acknowledgements
This interface was inspired by the DBAPI drivers for MySQL and Elasticsearch. Thanks.

- https://dev.mysql.com/doc/connector-python/en/connector-python-api-mysqlconnection-time-zone.html
- https://github.com/preset-io/elasticsearch-dbapi#time-zone

### Synopsis
Use `time_zone` argument when creating a new connection. Here, use a `pytz.timezone` object.
```python
>>> import pytz
>>> from crate.client import connect

>>> connect("localhost:4200", time_zone=pytz.timezone("Australia/Sydney"))
>>> cursor = connection.cursor()
>>> cursor.execute(stmt)
>>> cursor.fetchone()
['foo', datetime.datetime(2022, 7, 19, 4, 10, 36, 758000, tzinfo=<DstTzInfo 'Australia/Sydney' AEST+10:00:00 STD>)]
```

Use `time_zone` argument when creating a new cursor. Here, use an UTC offset string.
```python
>>> from crate.client import connect

>>> connection = connect("localhost:4200")
>>> cursor = connection.cursor(time_zone="+0530")
>>> cursor.execute(stmt)
>>> cursor.fetchone()
['foo', datetime.datetime(2022, 7, 18, 23, 40, 36, 758000, tzinfo=datetime.timezone(datetime.timedelta(seconds=19800), '+0530'))]
```

### References
- #395
- #426 
- #437 
- #442

### Backlog
- For bringing this functionality to the SQLAlchemy dialect, https://github.com/spoqa/sqlalchemy-utc may be of interest.

### DIFF
--- a/src/crate/client/connection.py
+++ b/src/crate/client/connection.py
@@ -47,6 +47,7 @@ class Connection(object):
                  socket_tcp_keepintvl=None,
                  socket_tcp_keepcnt=None,
                  converter=None,
+                 time_zone=None,
                  ):
         """
         :param servers:
@@ -103,9 +104,28 @@ class Connection(object):
         :param converter:
             (optional, defaults to ``None``)
             A `Converter` object to propagate to newly created `Cursor` objects.
+        :param time_zone:
+            (optional, defaults to ``None``)
+            A time zone specifier used for returning `TIMESTAMP` types as
+            timezone-aware native Python `datetime` objects.
+
+            Different data types are supported. Available options are:
+
+            - ``datetime.timezone.utc``
+            - ``datetime.timezone(datetime.timedelta(hours=7), name="MST")``
+            - ``pytz.timezone("Australia/Sydney")``
+            - ``zoneinfo.ZoneInfo("Australia/Sydney")``
+            - ``+0530`` (UTC offset in string format)
+
+            When `time_zone` is `None`, the returned `datetime` objects are
+            "naive", without any `tzinfo`, converted using ``datetime.utcfromtimestamp(...)``.
+
+            When `time_zone` is given, the returned `datetime` objects are "aware",
+            with `tzinfo` set, converted using ``datetime.fromtimestamp(..., tz=...)``.
         """
 
         self._converter = converter
+        self.time_zone = time_zone
 
         if client:
             self.client = client
@@ -135,10 +155,12 @@ class Connection(object):
         Return a new Cursor Object using the connection.
         """
         converter = kwargs.pop("converter", self._converter)
+        time_zone = kwargs.pop("time_zone", self.time_zone)
         if not self._closed:
             return Cursor(
                 connection=self,
                 converter=converter,
+                time_zone=time_zone,
             )
         else:
             raise ProgrammingError("Connection closed")
--- a/src/crate/client/cursor.py
+++ b/src/crate/client/cursor.py
@@ -18,8 +18,11 @@
 # However, if you have executed another commercial license agreement
 # with Crate these terms will supersede the license and you may use the
 # software solely pursuant to the terms of the relevant commercial agreement.
+from datetime import datetime, timedelta, timezone
 
+from .converter import DataType
 import warnings
+import typing as t
 
 from .converter import Converter
 from .exceptions import ProgrammingError
@@ -32,13 +35,15 @@ class Cursor(object):
     """
     lastrowid = None  # currently not supported
 
-    def __init__(self, connection, converter: Converter):
+    def __init__(self, connection, converter: Converter, **kwargs):
         self.arraysize = 1
         self.connection = connection
         self._converter = converter
         self._closed = False
         self._result = None
         self.rows = None
+        self._time_zone = None
+        self.time_zone = kwargs.get("time_zone")
 
     def execute(self, sql, parameters=None, bulk_parameters=None):
         """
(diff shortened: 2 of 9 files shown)


src/crate/client/cursor.py
```diff
@@ -238,3 +242,64 @@ def get_default_converter() -> Converter:
...
+        """
+        Convert UTC offset in string format (e.g. `+0530`) into `datetime.timezone` object.
+        """
+        assert len(tz) == 5, f"Time zone '{tz}' is given in invalid UTC offset format"
+        try:
+            hours = int(tz[:3])
+            minutes = int(tz[0] + tz[3:])
+            return timezone(timedelta(hours=hours, minutes=minutes), name=tz)
+        except Exception as ex:
+            raise ValueError(f"Time zone '{tz}' is given in invalid UTC offset format: {ex}")
```
amotl, line 305: I wonder if this helper function should gracefully also support the `+05:30` format, like it is used in the example for the [MySQL Connector/Python](https://github.com/mysql/mysql-connector-python) driver [^1].

[^1]: https://dev.mysql.com/doc/connector-python/en/connector-python-api-mysqlconnection-time-zone.html

matriv: Imho, I wouldn't bother to expand support to other possible formats, I think that documenting the acceptable formats is enough.

matriv (commented):
Thank you! It looks good, I have some comments around docs.

src/crate/client/doctests/cursor.txt
```diff
@@ -363,6 +363,60 @@ Proof that the converter works correctly, ``B\'0110\'`` should be converted to
     [6]
 
 
+``TIMESTAMP`` conversion with time zone
+=======================================
+
+Based on the data type converter machinery, the driver offers a convenient
```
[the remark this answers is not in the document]
amotl: Thanks. Fixed with 349b92e.

src/crate/client/connection.py
```diff
@@ -103,9 +104,14 @@ def __init__(self,
         :param converter:
             (optional, defaults to ``None``)
             A `Converter` object to propagate to newly created `Cursor` objects.
+        :param time_zone:
+            (optional, defaults to ``None``)
+            A time zone specifier used for returning `TIMESTAMP` types as
+            timezone-aware native Python `datetime` objects.
```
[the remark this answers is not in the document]
amotl: Thanks, cefa66a0ec improves the situation. Duplication probably not so easy avoidable.

src/crate/client/cursor.py
```diff
@@ -238,3 +242,64 @@ def get_default_converter() -> Converter:
...
+        return self._time_zone
+
+    @time_zone.setter
+    def time_zone(self, tz):
+        """
+        Set the time zone.
+
+        It supports different ways to populate. Some examples::
+
+        - ``datetime.timezone.utc``
```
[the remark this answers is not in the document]
amotl: Fixed with cefa66a0ec. Support for `zoneinfo.ZoneInfo` added with cad871c6ca. Thanks again.

src/crate/client/connection.py
```diff
@@ -103,9 +104,14 @@ def __init__(self,
         :param converter:
             (optional, defaults to ``None``)
             A `Converter` object to propagate to newly created `Cursor` objects.
+        :param time_zone:
+            (optional, defaults to ``None``)
```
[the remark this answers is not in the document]
amotl: When `time_zone` is not given, or `None`, the returned `datetime` objects will be naive, completely lacking any `tzinfo` information. I've also reflected it within cefa66a0ecac.

amotl (commented):
Dear Marios,

I've added some commits and hopefully addressed your suggestions well. Please let me know about anything missing, or where the patch could be improved further.

With kind regards,
Andreas.

src/crate/client/test_connection.py
```diff
@@ -23,7 +25,25 @@ def test_invalid_server_version(self):
         self.assertEqual((0, 0, 0), connection.lowest_server_version.version)
         connection.close()
 
-    def test_with_is_supported(self):
+    def test_context_manager(self):
```
matriv, line 28: `context_manager`? isn't `test_connection` more appropriate?

matriv (approved):
LGTM. thank you! Left a comment about a test method name.

src/crate/client/test_connection.py
amotl: Dear Marios,

the _whole module_ is about testing the `connect()` function in different variants. This specific case exercises the use of the `connect()` method as a context manager.

With kind regards,
Andreas.

matriv: Thx for clarifying!

src/crate/client/test_cursor.py
```diff
@@ -19,11 +19,14 @@
 # with Crate these terms will supersede the license and you may use the
 # software solely pursuant to the terms of the relevant commercial agreement.
 
-from datetime import datetime
+import datetime
+import zoneinfo
```
amotl, line 23: `zoneinfo` is only available in Python 3.9 and newer, see https://docs.python.org/3/library/zoneinfo.html.

```
ModuleNotFoundError: No module named 'zoneinfo'
```

-- https://github.com/crate/crate-python/actions/runs/3276036472/jobs/5391606304#step:4:212

amotl: The right solution is to probably use a backport package, see [`backports.zoneinfo`](https://pypi.org/project/backports.zoneinfo/).

src/crate/client/cursor.py
amotl: It is now documented here:

> For the time_zone keyword argument, different data types are supported. The available options are:

    [...]
    +0530 (UTC offset in string format)

-- https://github.com/crate/crate-python/blob/98630d0167b0f7802586fb7d3ffd7f25087d27ba/docs/query.rst#timestamp-conversion-with-time-zone

src/crate/client/connection.py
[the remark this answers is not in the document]
amotl: The details have been added to the function signature documentation docstring.

https://github.com/crate/crate-python/blob/98630d0167b0f7802586fb7d3ffd7f25087d27ba/src/crate/client/connection.py#L120-L124

hammerhead (approved):

PR #457, opened 2022-10-06
CI: `setup_ci` should run before `bootstrap.sh`

amotl:
On the "nightly" runs, the test suite did not actually use CrateDB nightly builds, but used the version defined in `versions.cfg` instead.

Because of that, it is currently using CrateDB 5.0.1, even on the nightly runs, see https://github.com/crate/crate-python/actions/runs/3194158943/jobs/5213450280#step:4:170.

### DIFF
--- a/.github/workflows/nightly.yml
+++ b/.github/workflows/nightly.yml
@@ -36,11 +36,11 @@ jobs:
       - name: Invoke tests
         run: |
 
-          # Bootstrap environment.
-          source bootstrap.sh
-
           # Propagate build matrix information.
           ./devtools/setup_ci.sh
+
+          # Bootstrap environment.
+          source bootstrap.sh
 
           # Invoke validation tasks.
           flake8 src bin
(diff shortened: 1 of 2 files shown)


seut (approved):
👍 good catch!

BaurzhanSakhariev (approved):

amotl:
If you slap a `workflow_dispatch` on your GHA workflow definition, like https://github.com/crate/crate-python/blob/b613f0dae84a94d83203049abfc72950d5727ebb/.github/workflows/nightly.yml#L1-L6 you can run the workflow on any branch you like on the corresponding page [1], which is sweet.

image

And indeed, it fails [2], not able to reflect the `PrimaryKeyConstraint()` 🌧️.

[1] https://github.com/crate/crate-python/actions/workflows/nightly.yml
[2] https://github.com/crate/crate-python/actions/runs/3198411459

PR #458, opened 2022-10-06
Dependencies: Limit `geojson` package to version <3

amotl:
Given that the package did not receive any update since 2019, chances are that the next release may bump to a major version, which might not be compatible.

### DIFF
--- a/setup.py
+++ b/setup.py
@@ -65,7 +65,7 @@ setup(
     },
     extras_require=dict(
         sqlalchemy=['sqlalchemy>=1.0,<1.5',
-                    'geojson>=2.5.0'],
+                    'geojson>=2.5.0,<3'],
         test=['tox>=3,<4',
               'zope.testing>=4,<5',
               'zope.testrunner>=5,<6',


matriv:
Isn't it better to stumble upon this upgrade, and decide then if we try to accommodate, or limit the versions?

amotl:
When a package manager supports pinning package versions using package lock files, you are usually describing similar things these days when using [caret requirements](https://python-poetry.org/docs/dependency-specification/#caret-requirements), like specifying `^1.2.3`, essentially expressing the version constraint `>=1.2.3 <2.0.0`, i.e. »allow pulling of minor and patch releases, but don't upgrade to the next major version« in the context of [SemVer](https://semver.org/) versioning.

On all projects I am touching recently, I try to adjust things that way, even when not able to use the »caret requirements« version specification.

> Isn't it better to stumble upon this upgrade.

I think it is not a good idea to let _users_ stumble into the upgrade.

matriv:
Thank you very much for the detailed description @amotl.
I was thinking that if the upgrade is pulled in, our nightly run for example will fail, and therefore we will then decide if we fix it or restrict the version dependencies.

amotl:
> I was thinking that if the upgrade is pulled in, our nightly run for example will fail.

It would fail at the same time for all other users and we would not be able to do anything about it. The intention of this patch is to prevent that, by taking care about this scenario in advance.

matriv:
Thx, It's very clear! thought it would just break the nightly "package".

PR #475, opened 2022-12-02
fix zsh globbing in bootstrap.sh

proddata:
Not sure, if this is relevant for all platforms, however took me some time to find that (again) on a mac with zsh.


amotl (approved):
Thank you!

PR #476, opened 2022-12-02
Release 0.28

hammerhead:
To get computed columns out to the customer requesting it.

## Checklist

### DIFF
--- a/CHANGES.txt
+++ b/CHANGES.txt
@@ -4,6 +4,11 @@ Changes for crate
 
 Unreleased
 ==========
+
+
+2022/12/02 0.28.0
+=================
+
 - Added a generic data type converter to the ``Cursor`` object, for converting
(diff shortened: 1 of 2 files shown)


setup.py
```diff
@@ -76,7 +76,7 @@ def read(path):
         doc=['sphinx>=3,<4',
              'crate-docs-theme'],
     ),
-    python_requires='>=3.4',
+    python_requires='>=3.7',
```
amotl, line 79: We decided to still support Python 3.6, when possible. See https://github.com/crate/crash/pull/369.

hammerhead: Was seeing that only Python >= 3.7 is mentioned in `classifiers`, but fine to keep it at 3.4.

amotl (approved):
👍

PR #477, opened 2022-12-02
Document `__table args__`

hammerhead:
As per the title.

## Checklist

### DIFF
--- a/setup.py
+++ b/setup.py
@@ -73,7 +73,7 @@ setup(
               'createcoverage>=1,<2',
               'stopit>=1.1.2,<2',
               'flake8>=4,<5'],
-        doc=['sphinx>=3,<4',
+        doc=['sphinx>=3.5,<6',
              'crate-docs-theme'],
     ),
     python_requires='>=3.4',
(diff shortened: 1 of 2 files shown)


setup.py
```diff
@@ -73,7 +73,7 @@ def read(path):
               'createcoverage>=1,<2',
               'stopit>=1.1.2,<2',
               'flake8>=4,<5'],
-        doc=['sphinx>=3,<4',
+        doc=['sphinx>=3.5,<6',
```
hammerhead, line 76: Python 3.10 support was added in Sphinx 4.3.0, so Sphinx in this project didn't run for me. This updates the version constraint to be aligned with the latest crate-docs-theme:

https://github.com/crate/crate-docs-theme/blob/66b054f08d0eab4c809a86a420c707c56dd7271e/setup.py#L61

amotl: I am fine with relaxing the upper bound. What about this change?
```suggestion
        doc=['sphinx>=3,<6',
```

amotl: Oh, same as `crate-docs-theme`. All right!

amotl (approved):
Thank you!

PR #479, opened 2022-12-06
Dependencies: Limit to `urllib3>=1.9,<2`, because urllib3-2.x is coming

amotl:
This patch is in the same spirit like #458.

Because urllib3 version 2.0 will be published soon [^1], it will be more safe to limit to urllib3 1.x, and to explicitly unlock it when we are sure the code base is compatible. #480 will exercise the update to the most recent alpha release of urllib3 2.x.

[^1]: See also https://pypi.org/project/urllib3/#history.

### DIFF
--- a/setup.py
+++ b/setup.py
@@ -24,11 +24,6 @@ import os
 import re
 
 
-requirements = [
-    'urllib3>=1.9',
-]
-
-
 def read(path):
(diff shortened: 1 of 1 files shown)


matriv (commented):
LGTM. Why not leave it as a set of requirements in case something else is added in the future?

hramezani (approved):
LGTM :+1:

Issue #483, opened 2022-12-09
SA20: Resolve all deprecation warnings using `SQLALCHEMY_WARN_20=1`


amotl:
### About
When running the test suite, it triggers a few deprecation warnings:
```
<stdin>:1: MovedIn20Warning: The ``declarative_base()`` function is now available as sqlalchemy.orm.declarative_base(). (deprecated since: 1.4) (Background on SQLAlchemy 2.0 at: https://sqlalche.me/e/b8d9)
<stdin>:1: RemovedIn20Warning: The ``bind`` argument to declarative_base is deprecated and will be removed in SQLAlchemy 2.0. (Background on SQLAlchemy 2.0 at: https://sqlalche.me/e/b8d9)
```
> 
> I was following the steps described here, but on the following `connection.execute` I get a `crate.client.exceptions.ProgrammingError: RelationUnknown[Relation 'locations' unknown]`. Is there a step missing to run the table creation? And how does it later return a row count of 13, where are these rows coming from?

### References
- GH-464
- GH-465

_Originally posted by @hammerhead in https://github.com/crate/crate-python/pull/464#discussion_r1044165573_

amotl:
I think I had already made a start on this at a single spot when looking at SA20 compatibility. See db3831a05480.

amotl:
Resolved with #485.

Issue #465, opened 2022-10-13
SA20: Evaluate compatibility with SQLAlchemy 2.0


amotl:
#### SQLAlchemy 2.0.0b1 Released

> The 2.0 series represents over three years of effort towards updating SQLAlchemy's usage model and architectures for modern Python use. The update's origins lie within the three major Python paradigms that came into existence well after SQLAlchemy's first releases in 2006: Python 3, pep-484 typing support, and asyncio. It's also in response to the changing nature of the Python programming community, vastly larger and home to many new developers with higher standards for strictness, ease of use and documentation than was the case sixteen years ago.
>
> [...]
>
> SQLAlchemy 2.0 also finalizes a very large number of API changes first introduced in 1.4. As such, even though SQLAlchemy 1.4 provided a [very comprehensive](https://docs.sqlalchemy.org/en/20/changelog/migration_20.html) upgrade path, we expect there to be lots of issues within the beginning of the series, both as projects accommodate the changes as well as for users of the new features, in particular the Declarative annotation support. We therefore hope that the beta series will have lots of testing done, and we will likely move from beta releases into release candidates as well, anticipating a 2.0 final release after some months.
>
> -- https://www.sqlalchemy.org/blog/2022/10/13/sqlalchemy-2.0.0b1-released/

#### What's new?, migration guide, changelog, and release notes
- https://docs.sqlalchemy.org/en/20/changelog/whatsnew_20.html
- https://docs.sqlalchemy.org/en/20/changelog/migration_20.html
- https://docs.sqlalchemy.org/en/20/changelog/changelog_20.html
- https://github.com/sqlalchemy/sqlalchemy/releases/tag/rel_2_0_0b1

#### On PyPI
> ![image](https://user-images.githubusercontent.com/453543/195658082-be169cd4-986f-4396-8ed4-fec96783e5d0.png)
>
> -- https://pypi.org/project/SQLAlchemy/#history

amotl:
After making sure we are on Python 3, the suggested second step of the migration guide is to run the test suite with `RemovedIn20Warnings` enabled.

```
SQLALCHEMY_WARN_20=1 ./bin/test
```

Here we go. Lot's of stuff to work on.

-- https://docs.sqlalchemy.org/en/20/changelog/migration_20.html#migration-to-2-0-step-two-turn-on-removedin20warnings

amotl:
[...]

PR #485, opened 2022-12-20
Towards compatibility with SQLAlchemy 2.x

amotl:
### About
Based on information from the migration guide referenced below, this patch brings the code base a bit closer to SQLAlchemy 2.0.

-- [SQLAlchemy migration to 2.0 Step Two - Turn on RemovedIn20Warnings](https://docs.sqlalchemy.org/en/20/changelog/migration_20.html#migration-to-2-0-step-two-turn-on-removedin20warnings)

### Details

- Declarative becomes a first class API, so it is now located at `sqlalchemy.orm.declarative_base`.
- The `Session` no longer supports "bound metadata" when it resolves the engine to be used for connectivity. This means that an Engine object **must** be passed to the constructor now.
- `connection.execute` will now only accept SQL statements wrapped in `text()`.
- When needing to run a driver-level SQL statement, the method `connection.exec_driver_sql` must be used.
- Library-level "Autocommit" aka. "implicit commit" removed from both Core and ORM. It is advised to use a context manager instead, like `with engine.begin() as conn:`.
- `select()` no longer accepts varied constructor arguments, only the "generative" style of `select()` will be supported. The list of columns / tables to select from should be passed positionally.
- The `autoload` option to `sa.MetaData` has been deprecated. `autoload_with` takes place, it is already there and supported.

### References
- GH-465
- GH-483
- [SQLAlchemy 2.0 - Major Migration Guide](https://docs.sqlalchemy.org/en/20/changelog/migration_20.html)

### Notes
df3d04c1d4 is a bit unrelated to the changes regarding SQLAlchemy. However, those improvements, finally resolving a few leftover socket leaks in the test suite, have been made _while_ working on mitigating the SA warnings, so that's why it's here. I think it will not harm the review process too much.

### Compatibility
With GH-497, this patch also retains compatibility with SQLAlchemy 1.3.


src/crate/client/sqlalchemy/tests/dict_test.py
```diff
@@ -41,7 +40,8 @@ class SqlAlchemyDictTypeTest(TestCase):
 
     def setUp(self):
         self.engine = sa.create_engine('crate://')
-        metadata = sa.MetaData()
+        # FIXME: Deprecated with SA20.
+        metadata = sa.MetaData(bind=self.engine)
```
amotl, line 44: This is the only spot where the test suite still croaks when removing the `bind` argument. Still, I would like to bring in this patch, in order to have a common ground for further activities, maybe together with @hammerhead.

src/crate/client/sqlalchemy/tests/insert_from_select_test.py
```diff
@@ -64,19 +63,20 @@ class CharacterArchive(Base):
...
 
-        sel = select([self.character.name, self.character.age]).where(self.character.status == "Archived")
+        sel = select(self.character.name, self.character.age).where(self.character.status == "Archived")
         ins = insert(self.character_archived).from_select(['name', 'age'], sel)
         self.session.execute(ins)
         self.session.commit()
+        # TODO: Verify if this is correct.
         self.assertSQL(
-            "INSERT INTO characters_archive (name, age) SELECT characters.name, characters.age FROM characters WHERE characters.status = ?",
+            "INSERT INTO characters_archive (name, age) SELECT characters.name, characters.age FROM characters WHERE characters.status = :status_1",
```
amotl, line 80: I believe this change is legitimate, and is coming from the new [Transparent SQL Compilation Caching](https://docs.sqlalchemy.org/en/20/changelog/migration_14.html#transparent-sql-compilation-caching-added-to-all-dql-dml-statements-in-core-orm) feature added with SQLAlchemy 1.4.

However, I still would like to validate the functionality beyond this test, to make sure my assumption is correct.

amotl: I would like to contradict my statement. 00384515 brings back the "qmark" paramstyle used by the CrateDB Python driver. For more details, see GH-490.

src/crate/client/sqlalchemy/__init__.py
```diff
@@ -19,7 +19,13 @@
 # with Crate these terms will supersede the license and you may use the
 # software solely pursuant to the terms of the relevant commercial agreement.
 
+from .compat.api13 import monkeypatch_add_exec_driver_sql
 from .dialect import CrateDialect
+from .sa_version import SA_1_4, SA_VERSION
+
+# SQLAlchemy 1.3 does not have the `exec_driver_sql` method.
+if SA_VERSION < SA_1_4:
+    monkeypatch_add_exec_driver_sql()
```
hammerhead, line 28: Do we already know how much longer we intend to support SQLAlchemy 1.3? According to SQLAlchemy's [version policy](https://www.sqlalchemy.org/download.html#status-desc-maintenance), 1.3 will become EOL as soon as 2.0 becomes stable. Given that maintaining compatibility with 1.3 here reaches a certain complexity, do we want to keep it much longer?

amotl: Hi. The most recent discussion about EOL of SA13 can be found at https://github.com/sqlalchemy/sqlalchemy/issues/8631#issuecomment-1329264864. I think we will support it for a while, and I am happy that the changes have not been too much intrusive, other than needing a few lines in `compat/api13.py`.

We discussed different alternatives on maintenance of the code within this package with @seut, regarding version deprecations and such, but all other alternatives would need more efforts, especially if we would need to deprecate support for SA13 earlier, because this might mean we would have to keep a maintenance branch or such. See also GH-494.

amotl: Nevertheless, I think it will not hurt to add a runtime warning for the next release, which will give users a notice that dropping support for SA13 will be around the corner.

hammerhead: The 1.3 support is quite nicely encapsulated, indeed. Well done! For now (while 1.3 is still supported upstream) it's a good solution. Once 1.3 reached EOL, I would still vote for removing it shortly after to eliminate the need for monkey patching (which always feels a bit risky to me, especially in tests).

hammerhead (approved):

src/crate/client/sqlalchemy/__init__.py
mfussenegger: +1 for removing 1.3 support. Given that crate-python is mostly stable (bugfixes are rare) we can point users of SQLAlchemy 1.3 to an older version of crate-python and make sure to update the version constraints for the next crate-python release. `pip` should then pick the right version, no?

amotl: I agree. See also GH-501.

PR #486, opened 2022-12-20
SQLAlchemy deprecation in `SqlAlchemyDictTypeTest`

hammerhead:
@amotl, this is an attempt at fixing the deprecation warning in `SqlAlchemyDictTypeTest`. I'm not 100% sure about the `test_update_with_dict_column`, this would need manual verification.

## Checklist

### DIFF
--- a/src/crate/client/sqlalchemy/tests/dict_test.py
+++ b/src/crate/client/sqlalchemy/tests/dict_test.py
@@ -37,14 +37,14 @@ FakeCursor.return_value = fake_cursor
 
 
 class SqlAlchemyDictTypeTest(TestCase):
-
     def setUp(self):
         self.engine = sa.create_engine('crate://')
-        # FIXME: Deprecated with SA20.
-        metadata = sa.MetaData(bind=self.engine)
-        self.mytable = sa.Table('mytable', metadata,
+        metadata = sa.MetaData()
+        self.mytable = sa.Table('mytable',
+                                metadata,
                                 sa.Column('name', sa.String),
-                                sa.Column('data', Craty))
+                                sa.Column('data', Craty),
+                                autoload_with=self.engine)
 
     def assertSQL(self, expected_str, actual_expr):
         self.assertEqual(expected_str, str(actual_expr).replace('\n', ''))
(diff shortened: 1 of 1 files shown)


src/crate/client/sqlalchemy/tests/dict_test.py
```diff
@@ -90,19 +91,18 @@ def test_select_with_dict_column_where_clause_other_col(self):
...
         )
 
     def test_update_with_dict_column(self):
         mytable = self.mytable
         stmt = mytable.update().\
             where(mytable.c.name == 'Arthur Dent').\
-            values({
-                "data['x']": "Trillian"
-            })
+            values({ mytable.c.data['x']: "Trillian" })
```
hammerhead, line 102: This looks to be valid:
```python
from crate.client.sqlalchemy import types
from sqlalchemy.orm import Session
import sqlalchemy as sa

engine = sa.create_engine("crate://")
metadata_obj = sa.MetaData()

mytable = sa.Table(
    "characters",
    metadata_obj,
    sa.Column("id", sa.String, primary_key=True),
    sa.Column("details", types.Object),
    autoload_with=engine,
)

session = Session(bind=engine)
session.execute(
    mytable.update()
    .where(mytable.c.id == "1")
    .values({mytable.c.details["x"]: "Trillian"})
)
```

Resulting query as per `sys.jobs_log`: `UPDATE characters SET details['x'] = ? WHERE characters.id = ?`

hammerhead: The previous test might have been wrong, using `values({"data['x']": "Trillian"})` throws an exception in a real-world example without any mocking: `TypeError: Object of type BindParameter is not JSON serializable`

amotl: Oh. Thanks for spotting and investigating diligently.

amotl: > throws an exception in a real-world example without any mocking

So, if that did not work before anyway, we will not have to announce anything as "breaking change", right? When looking at this again, I bet this was the spot which may have tripped me in the past already. Thanks again!

amotl: I've added 2c6db67c to make the tests succeed.

PR #487, opened 2022-12-20
Documentation: Improve wording, linking, and formatting

amotl:
Some copy editing across the board, with a focus on wording, linking, and formatting.


docs/sqlalchemy.rst
```diff
@@ -330,44 +329,48 @@ characters::
...
+    sub-columns.
 
 .. _objectarray:
 
 ``ObjectArray``
 ...............
 
 In addition to the `Object`_ type, the CrateDB SQLAlchemy dialect also provides
-a ``ObjectArray`` type, which is structured as a `list`_ of `dictionaries`_.
+a ``ObjectArray`` type, which is structured as a :class:`py:list` of
```
hammerhead, line 351: Very nice to have those direct links! Do you know why one renders in bold and the other one doesn't?

Screenshot 2022-12-27 at 13 02 01

amotl: That's weird, thanks for spotting. I do not have an idea yet, why this is happening.

amotl: ### Observations I
```html
<a class="reference external" href="https://docs.python.org/3/library/stdtypes.html#list" title="(in Python v3.11)">
  <code class="docutils literal notranslate">
    <span class="pre">list</span>
  </code>
</a>
```
```html
<a class="reference external" href="https://docs.python.org/3/library/stdtypes.html#dict" title="(in Python v3.11)">
  <code class="xref py py-class docutils literal notranslate">
    <span class="pre">dictionaries</span>
  </code>
</a>
```

It looks like the `xref` class (from `basic.css`) adds the `font-weight: bold;` CSS directive.


### Observations II
Funny enough, when rendering locally (not on RTD), those additional `code` tags, including the corresponding classes, are not rendered at all. Instead, the HTML looks like this:
```html
<a class="reference external" href="https://docs.python.org/3/library/stdtypes.html#lists">list</a>
<a class="reference external" href="https://docs.python.org/3/library/stdtypes.html#dict">dictionaries</a>
```

And visually, it looks like this:
image


### Next steps
So, I guess we will need to find out why the HTML output is different when being built on/by RTD.

amotl, line 351:
```suggestion
an ``ObjectArray`` type, which is structured as a :class:`py:list` of
```

amotl: After resetting my local sandbox environment, I am also getting the same results as above, where the rendered HTML at this spot includes `<code>` tags, and where the same CSS classes `xref py py-class` are present on the `:class:dictionaries <py:dict>` link.

It looks like the custom label, `dictionaries`, is responsible for the additonal `xref py py-class` classes.

amotl: I have not been able to discover the root cause, so I fixed it with https://github.com/crate/crate-docs-theme/pull/366.

amotl: With the patch to `crate-docs-theme`, it now looks like this:

image

hammerhead (approved):

PR #488, opened 2022-12-21
Add support for SQLAlchemy 2.0

amotl:
Based on GH-485 and GH-498, this patch concludes compatibility work with SQLAlchemy 2.0 [^1][^2] (GH-465) 🎉.

[^1]: https://docs.sqlalchemy.org/en/20/changelog/whatsnew_20.html
[^2]: https://docs.sqlalchemy.org/en/20/changelog/migration_20.html

### DIFF
--- a/bootstrap.sh
+++ b/bootstrap.sh
@@ -70,7 +70,7 @@ function setup_package() {
       if [ "${SQLALCHEMY_VERSION}" = "latest" ]; then
         pip install "sqlalchemy" --upgrade
       else
-        pip install "sqlalchemy==${SQLALCHEMY_VERSION}"
+        pip install "sqlalchemy==${SQLALCHEMY_VERSION}" --pre
       fi
     fi
 
--- a/bootstrap.sh
+++ b/bootstrap.sh
@@ -70,7 +70,7 @@ function setup_package() {
       if [ "${SQLALCHEMY_VERSION}" = "latest" ]; then
         pip install "sqlalchemy" --upgrade
       else
-        pip install "sqlalchemy==${SQLALCHEMY_VERSION}" --pre
+        pip install "sqlalchemy==${SQLALCHEMY_VERSION}"
       fi
     fi
 
--- a/src/crate/client/sqlalchemy/compat/core20.py
+++ b/src/crate/client/sqlalchemy/compat/core20.py
@@ -21,8 +21,9 @@
 
 from typing import Any, Dict, List, MutableMapping, Optional, Tuple, Union
 
+import sqlalchemy as sa
 from sqlalchemy import ColumnClause, ValuesBase, cast, exc
-from sqlalchemy.sql import crud, dml
+from sqlalchemy.sql import dml
 from sqlalchemy.sql.base import _from_objects
 from sqlalchemy.sql.compiler import SQLCompiler
 from sqlalchemy.sql.crud import (REQUIRED, _as_dml_column, _create_bind_param,
@@ -46,6 +47,11 @@ class CrateCompilerSA20(CrateCompiler):
             update_stmt, self, **kw
         )
         update_stmt = compile_state.statement
+
+        # [20] CrateDB patch.
+        if not compile_state._dict_parameters and \
+                not hasattr(update_stmt, '_crate_specific'):
+            return super().visit_update(update_stmt, **kw)
 
         toplevel = not self.stack
         if toplevel:
@@ -87,7 +93,8 @@ class CrateCompilerSA20(CrateCompiler):
         table_text = self.update_tables_clause(
             update_stmt, update_stmt.table, render_extra_froms, **kw
         )
-        crud_params_struct = crud._get_crud_params(
+        # [20] CrateDB patch.
+        crud_params_struct = _get_crud_params(
             self, update_stmt, compile_state, toplevel, **kw
         )
         crud_params = crud_params_struct.single_params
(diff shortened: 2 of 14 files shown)


src/crate/client/sqlalchemy/dialect.py
```diff
@@ -158,7 +159,11 @@ def process(value):
 class CrateDialect(default.DefaultDialect):
     name = 'crate'
     driver = 'crate-python'
-    statement_compiler = CrateCompiler
+    if SA_VERSION >= SA_2_0:
+        from .sa20 import CrateCompilerSA20
+        statement_compiler = CrateCompilerSA20
+    else:
+        statement_compiler = CrateCompiler
```
amotl, line 166: I will apply this kind of modularization to the head branch first, to split specializations for SA14 into a separate `sa14.py` module, respectively a `CrateCompilerSA14` class, beforehand. After that, this patch will fit in more naturally.

src/crate/client/sqlalchemy/tests/bulk_test.py
```diff
@@ -74,4 +74,4 @@ def test_bulk_save(self):
             ('Banshee', 26),
             ('Callisto', 37)
         )
-        self.assertEqual(expected_bulk_args, bulk_args)
+        self.assertSequenceEqual(expected_bulk_args, bulk_args)
```
amotl, line 77: With SA<2.0, `bulk_args` is a `tuple`, so `expected_bulk_args` was written as such. Now, with SA>=2.0, it is a `list`. Because this only concerns the call arguments to `cursor.executemany()`, which is invoked by SQLAlchemy, I think we do not need to put any kind of admonition notice about a breaking change into the changelog, as it looks like it will not concern the user of the library.

Please let me know if you think differently.

src/crate/client/sqlalchemy/tests/compiler_test.py
```diff
@@ -74,25 +74,22 @@ def test_select_with_offset(self):
         """
         Verify the `CrateCompiler.limit_clause` method, with offset.
         """
-        self.metadata.bind = self.crate_engine
         selectable = self.mytable.select().offset(5)
-        statement = str(selectable.compile())
+        statement = str(selectable.compile(bind=self.crate_engine))
```
amotl, line 78: Well, that's the same thing what I elaborated about at GH-490.

> "Implicit" and "Connectionless" execution, and "bound metadata", have been removed beginning with SQLAlchemy 2.0 [^1].

In this case, the use of "bound metadata" was not caught by turning on [`RemovedIn20Warnings`](https://docs.sqlalchemy.org/en/20/changelog/migration_20.html#migration-to-2-0-step-two-turn-on-removedin20warnings) within the preparation step GH-485, probably because the code is _directly_ setting the attribute(s) on the _existing_ instances here, without running `bind=` through any other regular constructor or initializer methods.

[^1]: https://docs.sqlalchemy.org/en/20/changelog/migration_20.html#implicit-and-connectionless-execution-bound-metadata-removed

src/crate/client/sqlalchemy/dialect.py
amotl: > I will apply this kind of modularization to the head branch first.

See GH-498.

src/crate/client/sqlalchemy/tests/compiler_test.py
amotl: Diverted to GH-485.

src/crate/client/sqlalchemy/tests/dict_test.py
```diff
@@ -44,7 +44,8 @@
 
     def setUp(self):
         self.engine = sa.create_engine('crate://')
-        metadata = sa.MetaData()
+        # FIXME: Deprecated with SA20.
+        metadata = sa.MetaData(bind=self.engine)
```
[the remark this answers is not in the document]
amotl: I think this spot has been fixed, but currently there is no way to hide this admonition. See also https://github.com/github/codeql-action/issues/1411.

src/crate/client/sqlalchemy/compat/core20.py
```diff
@@ -0,0 +1,459 @@
...
+            if isinstance(k, str) and '[' in k:
+                bindparam = sa.sql.bindparam(k, v)
+                clause = k + ' = ' + self.process(bindparam)
+                set_clauses.append(clause)
+
+        text += ', '.join(set_clauses)
+        # [20] CrateDB patch end.
+
+        """
+        # TODO: Complete SA20 migration.
```
mfussenegger, line 137: Shouldn't this TODO get addressed with this PR?

src/crate/client/sqlalchemy/compat/core20.py
```diff
@@ -0,0 +1,459 @@
...
+    # It can be quickly reproduced by activating this section and running the
+    # test cases::
+    #
+    #   ./bin/test -vvvv -t dict_test
+    #
+    # That croaks like::
+    #
+    #   sqlalchemy.exc.CompileError: Unconsumed column names: characters_name
+    #
+    # TODO: Investigate why this is actually happening and eventually mitigate
```
mfussenegger, line 407: Same here

amotl:
Updating to the most recent SQLAlchemy 2.0 GA trips the test suite. It still worked with the previous 2.0.0rc3 release. See also GH-506.

src/crate/client/sqlalchemy/compat/core20.py
amotl: Hi. Thanks for spotting this leftover comment. I've spent another day on it, in order to mangle the code/data into a shape that would be accepted by the cast, but wasn't able to make a cut. So, I've just created GH-508 in order to track this issue, and eventually resolve it on behalf of a subsequent iteration.

src/crate/client/sqlalchemy/compat/core20.py
amotl: This is an issue we are carrying over since the initial implementation of the CrateDB SQLAlchemy dialect. I've also tried to remedy it a few times, but did not manage to improve it yet. At least, I've just created GH-509 in order to highlight this spot more prominently. I'd also recommend to investigate it on behalf on a subsequent iteration, in order not to block this patch.

bootstrap.sh
```diff
@@ -70,7 +70,7 @@ function setup_package() {
       if [ "${SQLALCHEMY_VERSION}" = "latest" ]; then
         pip install "sqlalchemy" --upgrade
       else
-        pip install "sqlalchemy==${SQLALCHEMY_VERSION}"
+        pip install "sqlalchemy==${SQLALCHEMY_VERSION}" --pre
```
mfussenegger, line 73: Is the `--pre` still required?

src/crate/client/sqlalchemy/sa_version.py
```diff
@@ -25,3 +25,4 @@
 SA_VERSION = V(sa.__version__)
 
 SA_1_4 = V('1.4.0b1')
+SA_2_0 = V('2.0.0b1')
```
mfussenegger, line 28:
```suggestion
SA_2_0 = V('2.0.0')
```

mfussenegger (approved):

bootstrap.sh
amotl: Thanks for spotting. I don't think so, so I've updated 71a7f0ad73 correspondingly.

src/crate/client/sqlalchemy/sa_version.py
amotl: Thanks. See also 71a7f0ad73.

PR #500, opened 2022-12-25
SQLAlchemy: Clean up `CrateCompiler`

amotl:
While working on the other refactorings and improvements, I discovered two spots that could be optimized.

### DIFF
--- a/src/crate/client/sqlalchemy/compiler.py
+++ b/src/crate/client/sqlalchemy/compiler.py
@@ -208,8 +208,6 @@ class CrateTypeCompiler(compiler.GenericTypeCompiler):
 
 class CrateCompiler(compiler.SQLCompiler):
 
-    prefetch = []
-
     def visit_getitem_binary(self, binary, operator, **kw):
         return "{0}['{1}']".format(
             self.process(binary.left, **kw),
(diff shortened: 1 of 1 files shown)


src/crate/client/sqlalchemy/compiler.py
```diff
@@ -208,8 +208,6 @@ def visit_ARRAY(self, type_, **kw):
 
 class CrateCompiler(compiler.SQLCompiler):
 
-    prefetch = []
```
amotl, line 211: This has been added with GH-188 without further reasoning. Originally, it is a property, which did not change much between SQLAlchemy 1.0 [^1] and 2.0 [^2].

Can you imagine any harm when removing it?

```python
@property
def prefetch(self):
    return list(self.insert_prefetch) + list(self.update_prefetch)
```

[^1]: https://github.com/sqlalchemy/sqlalchemy/blob/rel_1_1_0/lib/sqlalchemy/sql/compiler.py#L445-L447
[^2]: https://github.com/sqlalchemy/sqlalchemy/blob/rel_2_0_0b4/lib/sqlalchemy/sql/compiler.py#L1256-L1258

hammerhead: The oldest upstream commit I could find with references to `self.prefetch` was removing it before the SQLAlchemy 1.0 release: https://github.com/sqlalchemy/sqlalchemy/commit/3cc0cf0148ab0905e9903e282f7d1e1190d1c871. So looks safe to remove for me.

hammerhead (approved):

src/crate/client/sqlalchemy/compiler.py
amotl: Thank you!

Issue #501, opened 2023-01-09
Deprecation notice about support for SQLAlchemy 1.3


amotl:
Dear community,

at https://github.com/crate/crate-python/pull/485#discussion_r1057806508, it became clear we will remove support for SQLAlchemy 1.3 on the corresponding CrateDB dialect within the upcoming releases. It is effectively EOL, even more when SQLAlchemy 2.0 will be released, which is around the corner. Please raise your voice if there are any objections to be considered.

We are issuing this notice early, and will also add a corresponding `DeprecationWarning` to another upcoming maintenance release, in order to make the intentions clear, and to inform users properly.

With kind regards,
Andreas.

amotl:
A corresponding EOL/deprecation warning will be added with GH-503.

amotl:
`crate-0.30.0` has been released, and ships with the corresponding deprecation warning, still supporting all of SQLAlchemy 1.3, 1.4 and 2.0. If you are on 1.3, you should consider migrating to a higher version of SQLAlchemy soon. See also:

- [What’s New in SQLAlchemy 1.4]
- [What's New in SQLAlchemy 2.0]
- [SQLAlchemy 2.0 migration guide]

[What’s New in SQLAlchemy 1.4]: https://docs.sqlalchemy.org/en/20/changelog/migration_14.html
[SQLAlchemy 2.0 migration guide]: https://docs.sqlalchemy.org/en/20/changelog/migration_20.html
[What's New in SQLAlchemy 2.0]: https://docs.sqlalchemy.org/en/20/changelog/whatsnew_20.html

PR #503, opened 2023-01-26
QA: Add EOL/deprecation warning about support for SQLAlchemy 1.3

amotl:
### Introduction
We decided to drop support for SQLAlchemy 1.3 within one of the upcoming releases and are tracking this at GH-501.

### About
This patch adds a corresponding EOL/deprecation warning to the package, which will be emitted at runtime when running the Python interpreter with default warnings enabled [^enable-warnings][^option-W].

```console
python -Wd -c "import crate.client.sqlalchemy"
PYTHONWARNINGS=default python -c "import crate.client.sqlalchemy"
```

```text
/path/to/crate/client/sqlalchemy/__init__.py:38: DeprecationWarning: WARNING: SQLAlchemy 1.3 is effectively EOL.
With SQLAlchemy 2 around the corner, this will be even more immanent.
Future versions of the CrateDB SQLAlchemy dialect will drop support for SQLAlchemy 1.3.
It is recommended that you transition to using SQLAlchemy 1.4 or 2.0.
```

[^enable-warnings]: https://docs.python.org/3/library/warnings.html#updating-code-for-new-versions-of-dependencies
[^option-W]: https://docs.python.org/3/using/cmdline.html#cmdoption-W

### DIFF
--- a/src/crate/client/sqlalchemy/__init__.py
+++ b/src/crate/client/sqlalchemy/__init__.py
@@ -23,9 +23,27 @@ from .compat.api13 import monkeypatch_add_exec_driver_sql
 from .dialect import CrateDialect
 from .sa_version import SA_1_4, SA_VERSION
 
-# SQLAlchemy 1.3 does not have the `exec_driver_sql` method.
+
 if SA_VERSION < SA_1_4:
+    import textwrap
+    import warnings
+
+    # SQLAlchemy 1.3 is effectively EOL.
+    SA13_DEPRECATION_WARNING = textwrap.dedent("""
+    WARNING: SQLAlchemy 1.3 is effectively EOL.
+
+    SQLAlchemy 1.3 is EOL since 2023-01-27.
+    Future versions of the CrateDB SQLAlchemy dialect will drop support for SQLAlchemy 1.3.
+    It is recommended that you transition to using SQLAlchemy 1.4 or 2.0:
+
+    - https://docs.sqlalchemy.org/en/14/changelog/migration_14.html
+    - https://docs.sqlalchemy.org/en/20/changelog/migration_20.html
+    """.lstrip("\n"))
+    warnings.warn(message=SA13_DEPRECATION_WARNING, category=DeprecationWarning)
+
+    # SQLAlchemy 1.3 does not have the `exec_driver_sql` method, so add it.
     monkeypatch_add_exec_driver_sql()
+
 
 __all__ = [
     CrateDialect,
--- a/CHANGES.txt
+++ b/CHANGES.txt
@@ -5,6 +5,8 @@ Changes for crate
 Unreleased
 ==========
 
+- Add deprecation warning about dropping support for SQLAlchemy 1.3 soon, it is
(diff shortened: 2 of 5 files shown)


src/crate/client/sqlalchemy/__init__.py
```diff
@@ -23,10 +23,24 @@
...
 
-# SQLAlchemy 1.3 does not have the `exec_driver_sql` method.
+
 if SA_VERSION < SA_1_4:
+    import textwrap
+    import warnings
+
+    # SQLAlchemy 1.3 is effectively EOL.
+    SA13_DEPRECATION_WARNING = textwrap.dedent("""
+    WARNING: SQLAlchemy 1.3 is effectively EOL.
```
hlcianfagna, line 33:
```suggestion
    WARNING: SQLAlchemy 1.3 is currently in maintenance status and will soon be EOL.
```

src/crate/client/sqlalchemy/__init__.py
```diff
@@ -23,10 +23,24 @@
...
+    With SQLAlchemy 2 around the corner, this will be even more immanent.
```
hlcianfagna, line 34:
```suggestion
    With SQLAlchemy 2 around the corner, this will be even more imminent.
```

src/crate/client/sqlalchemy/__init__.py
amotl: I've adjusted the patch accordingly, thanks.

src/crate/client/sqlalchemy/__init__.py
amotl: Thanks for spotting this!

hlcianfagna (approved):

hlcianfagna:
@amotl please check if the change I made on https://github.com/crate/crate-python/pull/503/commits/2913ca94b6ee995477b9a3edf850e742f1baa8f9 makes sense

src/crate/client/sqlalchemy/__init__.py
```diff
@@ -23,10 +23,24 @@
...
+    import textwrap
+    import warnings
+
+    # SQLAlchemy 1.3 is effectively EOL.
+    SA13_DEPRECATION_WARNING = textwrap.dedent("""
+    WARNING: SQLAlchemy 1.3 is currently in maintenance status and will soon be EOL.
+    With SQLAlchemy 2 around the corner, this will be even more imminent.
+    Future versions of the CrateDB SQLAlchemy dialect will drop support for SQLAlchemy 1.3.
+    It is recommended that you transition to using SQLAlchemy 1.4 or 2.0.
+    """.lstrip("\n"))
```
hammerhead, line 37: 2.0 was released stable last night, and 1.3 is now officially EOL as per the table on https://www.sqlalchemy.org/download.html.

amotl: Thanks. I've adjusted the admonition text correspondingly.

hammerhead (approved):

PR #504, opened 2023-01-26
Maintenance: Update software versions for sandbox and CI

amotl:
Just a little maintenance patch to refresh software versions.


.github/workflows/tests.yml
```diff
@@ -21,8 +21,8 @@ jobs:
       matrix:
         os: ['ubuntu-latest', 'macos-latest']
         python-version: ['3.7', '3.8', '3.9', '3.10', '3.11']
-        cratedb-version: ['5.1.2']
-        sqla-version: ['1.3.24', '1.4.45']
+        cratedb-version: ['5.2.0']
```
hlcianfagna, line 24: Maybe 5.1.3 or 5.2.1 ? (I think a decision was made to not promote 5.2.0  to stable)

amotl: Thanks. I will update to CrateDB 5.2.1 right away when there will be a tarball.

bootstrap.sh
```diff
@@ -17,9 +17,9 @@
 # set -x
 
 # Default variables.
-BUILDOUT_VERSION=${BUILDOUT_VERSION:-2.13.7}
-CRATEDB_VERSION=${CRATEDB_VERSION:-5.1.2}
-SQLALCHEMY_VERSION=${SQLALCHEMY_VERSION:-1.4.45}
+BUILDOUT_VERSION=${BUILDOUT_VERSION:-2.13.8}
+CRATEDB_VERSION=${CRATEDB_VERSION:-5.2.0}
```
BaurzhanSakhariev, line 21: same as comment above, let's make it 5.2.1 when tarball is out

BaurzhanSakhariev (approved):

PR #510, opened 2023-02-09
CI: Add test matrix slot to test prerelease versions of Python package dependencies

amotl:
Because urllib3 version 2 is not released yet, this patch will allow to test it on CI, on behalf of another test matrix slot.


.github/workflows/tests.yml
```diff
@@ -33,10 +35,21 @@ jobs:
             python-version: '3.9'
           - os: 'macos-latest'
             python-version: '3.10'
+
+        # Another CI test matrix slot to test against prerelease versions of Python packages.
+        include:
+          - os: 'ubuntu-latest'
+            python-version: '3.11'
+            cratedb-version: '5.2.0'
```
matriv, line 43: `5.2.2`?

amotl: Version bumps will arrive with GH-511.

PR #511, opened 2023-02-09
CI: Dependency version bumps, and a few cleanups

amotl:
What the title says.

### DIFF
--- a/.github/workflows/codeql.yml
+++ b/.github/workflows/codeql.yml
@@ -30,7 +30,7 @@ jobs:
       fail-fast: false
       matrix:
         language: [ python ]
-        sqla-version: ['1.3.24', '1.4.46', '2.0.0']
+        sqla-version: ['<1.4', '<1.5', '<2.1']
 
     steps:
       - name: Checkout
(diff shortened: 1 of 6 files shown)


hlcianfagna (approved):

.github/workflows/codeql.yml
```diff
@@ -48,8 +48,8 @@ jobs:
 
       - name: Install project
         run: |
-          pip install --editable=.[sqlalchemy,test,doc]
-          pip install "sqlalchemy==${{ matrix.sqla-version }}" --upgrade --pre
+          pip install --editable=.[sqlalchemy,test]
```
matriv, line 51: just double checking, is removing `doc` correct?

matriv (approved):
LGTM, left a question

.github/workflows/codeql.yml
amotl: Thanks for asking. Yes, the `doc` extra had to be removed because its dependencies depressed `urllib3` to `<2`, rendering GH-510 useless.

The dependencies are not required on CI/GHA for those jobs, there is a separate workflow definition file [`docs.yml`](https://github.com/crate/crate-python/blob/master/.github/workflows/docs.yml) for that.

PR #513, opened 2023-02-09
Maintenance (PEP 632): Remove dependency on `distutils`

amotl:
### About
The `distutils` package will become deprecated, see [PEP 632 – Deprecate distutils module](https://peps.python.org/pep-0632/).

> DeprecationWarning: The distutils package is deprecated and slated for removal in Python 3.12. Use setuptools or check PEP 632 for potential alternatives

### Solution
The accepted solution is to vendor [`packaging.version`](https://github.com/pypa/packaging/blob/main/src/packaging/version.py), like how NumPy, pandas and SciPy are doing it equally.

### References
- https://github.com/numpy/numpy/issues/18588
- https://github.com/numpy/numpy/blob/v1.25.0.dev0/numpy/compat/_pep440.py
- https://github.com/pandas-dev/pandas/blob/v1.5.3/pandas/util/version/__init__.py
- https://github.com/scipy/scipy/blob/v1.10.0/scipy/_lib/_pep440.py

### DIFF
--- a/.coveragerc
+++ b/.coveragerc
@@ -4,8 +4,9 @@ omit =
     */.buildout/eggs/*
     bin/test
     buildout-cache/eggs/*
-    parts/*
     eggs/*
+    parts/*
+    src/crate/client/_pep440.py
 exclude_lines =
     # pragma: no cover
     raise NotImplemented
--- a/src/crate/client/_pep440.py
+++ b/src/crate/client/_pep440.py
@@ -1,0 +1,494 @@
+"""Utility to compare pep440 compatible version strings.
+
+The LooseVersion and StrictVersion classes that distutils provides don't
+work; they don't recognize anything like alpha/beta/rc/dev versions.
+
+This specific file has been vendored from NumPy on 2023-02-10 [1].
+Its reference location is in `packaging` [2,3].
+
+[1] https://github.com/numpy/numpy/blob/v1.25.0.dev0/numpy/compat/_pep440.py
+[2] https://github.com/pypa/packaging/blob/23.0/src/packaging/_structures.py
+[3] https://github.com/pypa/packaging/blob/23.0/src/packaging/version.py
+"""
+
+# Copyright (c) Donald Stufft and individual contributors.
+# All rights reserved.
+
+# Redistribution and use in source and binary forms, with or without
+# modification, are permitted provided that the following conditions are met:
+
+#     1. Redistributions of source code must retain the above copyright notice,
+#        this list of conditions and the following disclaimer.
+
+#     2. Redistributions in binary form must reproduce the above copyright
+#        notice, this list of conditions and the following disclaimer in the
+#        documentation and/or other materials provided with the distribution.
+
+# THIS SOFTWARE IS PROVIDED BY THE COPYRIGHT HOLDERS AND CONTRIBUTORS "AS IS"
+# AND ANY EXPRESS OR IMPLIED WARRANTIES, INCLUDING, BUT NOT LIMITED TO, THE
+# IMPLIED WARRANTIES OF MERCHANTABILITY AND FITNESS FOR A PARTICULAR PURPOSE
+# ARE DISCLAIMED. IN NO EVENT SHALL THE COPYRIGHT HOLDER OR CONTRIBUTORS BE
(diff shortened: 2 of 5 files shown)


src/crate/client/pep440.py
```diff
@@ -0,0 +1,586 @@
...
+        return self._key > other._key
+
+    def __ne__(self, other: object) -> bool:
+        if not isinstance(other, _BaseVersion):
+            return NotImplemented
+
+        return self._key != other._key
+
+
+class LegacyVersion(_BaseVersion):
```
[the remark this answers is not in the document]
amotl: Thanks for the admonition. In this case, `__eq__` is only derived from `self._key`.

src/crate/client/pep440.py
```diff
@@ -0,0 +1,582 @@
+# Vendored from `packaging` on 2021-30-04, via pandas.
+# - https://github.com/pypa/packaging/blob/ae891fd74/packaging/_structures.py
+# - https://github.com/pypa/packaging/blob/ae891fd74/packaging/version.py
+# - https://github.com/pandas-dev/pandas/blob/cf2b88b5ea/pandas/util/version/__init__.py
+
+# This file is dual licensed under the terms of the Apache License, Version
+# 2.0, and the BSD License. See the LICENSE file in the root of this repository
+# for complete details.
```
seut, line 8: If we use this license related header, we'd have to include the original license note to our LICENSE file afaik similar how pandas did it https://github.com/pandas-dev/pandas/commit/f40e58cc50abcd03efae372df0586467a957f695#diff-17d4b6290d4495e094bd6dc27902057dcb25ca7f73d595162899c22d1b95f43eR180. 

Or we keep the license header of the donald like numpy did https://github.com/numpy/numpy/pull/21000/files#diff-41a2b10dd4230acb81d1f9f7dedd8e8e6f6ef4f44d15f20105e9e82cf5a489ebR7.

seut (changes requested):

src/crate/client/pep440.py
amotl: ### Response
Thank you for your attention.

### Thoughts
I think it would be even better if someone would release and maintain a dedicated `versions`, or `pep440` package, that this code would not be vendored over and over again.

### Discovery
Better late than never: It turns out, that happened already!

- https://pypi.org/project/versions/
- https://pypi.org/project/pep440/

### Thoughts
Maybe better spend another iteration to re-use one of those packages instead?

amotl: > Maybe better spend another iteration to re-use one of those packages instead?

Hm. While `versions` feels a bit too heavy, I think `pep440` is a bit too thin.

seut: Yes sounds good, lets use one of them.

seut: > > Maybe better spend another iteration to re-use one of those packages instead?
> 
> Hm. While `versions` feels a bit too heavy, I think `pep440` is a bit too thin.

Another one: https://pypi.org/project/pep440-version-utils/
But also not sure if this is really worth of a dependency. Maybe we should just fix the license header instead?

amotl: > Another one: https://pypi.org/project/pep440-version-utils/.

Thanks, but this one actually _pulls in_ `packaging`, see [pyproject.toml#L16](https://github.com/m-vdb/pep440-version-utils/blob/10e1eeb8839c97b91442fe08c117ea02d278ba92/pyproject.toml#L16).

> But also not sure if this is really worth of a dependency.

Yeah, I think that was the same reasoning of the other maintainers.

> Maybe we should just fix the license header instead?

Sure, let's do it like that for now.

mfussenegger:
Wouldn't a `tuple(map(int, version.split('.')))` be good enough for our purposes? CrateDB versions follow x.y.z and don't have any other components.  A tuple with 3 integers has the correct comparison semantics

src/crate/client/pep440.py
amotl: I've reworked the patch to use the `_pep440.py` file from current NumPy [^1], as you suggested.

[^1]: https://github.com/numpy/numpy/blob/v1.25.0.dev0/numpy/compat/_pep440.py

seut (approved):

PR #514, opened 2023-02-09
SQLAlchemy: Don't use deprecated `_ConnectionFairy.connection` attribute

amotl:
### About

SQLAlchemy's `_ConnectionFairy.connection` attribute will be become deprecated.

> SADeprecationWarning: The _ConnectionFairy.connection attribute is deprecated; please use 'driver_connection' (deprecated since: 2.0)

### Solution
1. Just use `driver_connection` instead.
2. For SA13, addtionally back-port it.

### DIFF
--- a/src/crate/client/sqlalchemy/compat/api13.py
+++ b/src/crate/client/sqlalchemy/compat/api13.py
@@ -131,3 +131,26 @@ def monkeypatch_amend_select_sa14():
     sqlalchemy.select = select_sa14
     sqlalchemy.sql.select = select_sa14
     sqlalchemy.sql.expression.select = select_sa14
+
+
+@property
+def connectionfairy_driver_connection_sa14(self):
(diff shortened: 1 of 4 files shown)


hlcianfagna (approved):

src/crate/client/sqlalchemy/compat/api13.py
```diff
@@ -131,3 +131,26 @@ def monkeypatch_amend_select_sa14():
     sqlalchemy.select = select_sa14
     sqlalchemy.sql.select = select_sa14
     sqlalchemy.sql.expression.select = select_sa14
+
+
+@property
+def connectionfairy_driver_connection_sa14(self):
```
seut, line 137: Isn't this only added for SA < 1.4?
```suggestion
def connectionfairy_driver_connection_sa13(self):
```

seut (approved):
👍

PR #530, opened 2023-03-04
Chore: Adjust minimum Python version in project metadata

fabaff:
Align with classifiers and what's tested by CI.

## Checklist

### DIFF
--- a/setup.py
+++ b/setup.py
@@ -77,7 +77,7 @@ setup(
         doc=['sphinx>=3.5,<7',
              'crate-docs-theme>=0.26.5'],
     ),
-    python_requires='>=3.4',
+    python_requires='>=3.7',
     package_data={'': ['*.txt']},
     classifiers=[
         'Development Status :: 5 - Production/Stable',
--- a/setup.py
+++ b/setup.py
@@ -77,7 +77,7 @@ setup(
         doc=['sphinx>=3.5,<7',
              'crate-docs-theme>=0.26.5'],
     ),
-    python_requires='>=3.7',
+    python_requires='>=3.6',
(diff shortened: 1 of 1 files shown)


amotl:
Dear Fabian,

thank you very much for submitting this adjustment. I've just verified the package will also install on Python 3.6, but will not work on earlier versions, because we started using f-strings already. While Python 3.6 is officially end-of-life already, we don't want to be too restrictive to prevent installation on older Python installations.

However, it still makes sense to stop announcing it within the classifiers list, or shall we bring it back? WDYT?

With kind regards,
Andreas.

### Appendix
#### Python 3.5
```console
$ docker run --rm -it python:3.5 bash
$ pip install crate
$ python -c "import crate.client"
```
```python
  File "/usr/local/lib/python3.5/site-packages/crate/client/cursor.py", line 311
    assert len(tz) == 5, f"Time zone '{tz}' is given in invalid UTC offset format"
                                                                                 ^
SyntaxError: invalid syntax
```
#### Python 3.6
```console
$ docker run --rm -it python:3.6 bash
$ pip install crate[sqlalchemy]
$ python -c "import crate.client.sqlalchemy"
```

setup.py
```diff
@@ -77,7 +77,7 @@ def read(path):
         doc=['sphinx>=3.5,<7',
              'crate-docs-theme>=0.26.5'],
     ),
-    python_requires='>=3.4',
+    python_requires='>=3.7',
```
amotl, line 80:
```suggestion
    python_requires='>=3.6',
```

amotl (approved):

PR #464, opened 2022-10-13
Refactor certain doctests towards either reStructuredText documentation or Python tests

amotl:
### About
This is a stab at #448, to improve the shape of doctests vs. rendered documentation. It moves some doctest files around into the documentation directory, and dissolves others into pure-Python tests.

### New documentation section

The doctest files referenced below are now becoming first citizens of the public documentation, within a dedicated "by example" section. To make sure they are always valid, they are still evaluated as doctests. Many thanks to the authors who originally conceived them.

After adding some introductory paragraphs to the individual files, I think the new content is sweet already, and will be even better after another iteration of more content-related edits and refactorings by a subsequent patch.

Here, the new documentation section is rendered as a preview on RTD, see [CrateDB Python client by example](https://crate-python--464.org.readthedocs.build/en/464/by-example/).

### Details
- `doctests/{client,http,blob}.txt` and `doctests/{connection,cursor,https}.txt` have been moved to the documentation, to be rendered as reStructuredText (10da69c5f8, 1731cc558).
- All SQLAlchemy-related doctests are now also part of the documentation, that is `doctests/sqlalchemy.rst` and `sqlalchemy/doctests/{dialect,itests,reflection}.txt` (13ed96f2), which are now located at `docs/by-example/sqlalchemy/{cru,getting-started,inspection-reflection}.rst`.
- On the other hand, `doctests/layer.txt` and `doctests/mocking.txt` are now implemented as regular Python tests.
- Any kinds of symbol aliases have been removed from the doctest launcher, established per `test.globs`, in order not to obfuscate them away from the users reading the examples, who are likely aiming to just copy/paste the presented code snippets.

PR #536, opened 2023-03-13
Documentation: Improve landing page and other sections across the board

amotl:
### About
Usability improvements in the same spirit as previous patches, this time not forgetting about [CrateDB Cloud](https://console.cratedb.cloud/).

- GH-443
- GH-464
- GH-487

### View
Other than reviewing the details on a diff/patch level, you may also want to visit the rendered documentation, in order to get an overall impression.

- Old: https://crate.io/docs/python/
- New: https://crate-python--536.org.readthedocs.build/en/536/

### Review
I know reviewing such a noisy diff is daunting. I appreciate your efforts in advance.


docs/index.rst
```diff
@@ -1,51 +1,211 @@
...
+    >>>     print(cursor.fetchall())
+    >>>     cursor.close()
+
+Connect to `CrateDB Cloud`_.
+
+.. code-block:: python
+
+    >>> # Connect using DB API.
+    >>> from crate import client
+    >>> client.connect("https://admin:iM_n1%0xRzShsapU4U(lDGOj@example.aks1.westeurope.azure.cratedb.net:4200")
```
hammerhead, line 81: Might be nicer to read by using the `username`/`password` parameters:
```python
client.connect("https://example.aks1.westeurope.azure.cratedb.net:4200", username="admin", password="iM_n1%0xRzShsapU4U(lDGOj")
```

docs/index.rst
```diff
@@ -1,51 +1,211 @@
...
+.. toctree::
+    :maxdepth: 2
+
+    sqlalchemy
+
+Install package from PyPI with DB API and SQLAlchemy support.
+
+.. code-block:: shell
+
+    $ pip install crate[sqlalchemy] pandas
```
hammerhead, line 100: Depending on the shell, it might need to be wrapped in double quotes (for example when using zsh): `pip install "crate[sqlalchemy]" pandas`. This syntax is also used in `README.rst` (otherwise, installation will error with `zsh: no matches found: crate[sqlalchemy]`).

amotl: Thanks for spotting 💯. @proddata told me he actually needed to use single quotes on zsh instead?

hammerhead: Both single and double quotes work fine for me (zsh on macOS). Not sure if one is more correct than the other 🤷.

docs/index.rst
```diff
@@ -1,51 +1,211 @@
...
 
-    This is a basic CrateDB driver reference.
+The Python client library for `CrateDB`_ implements the Python Database API
+Specification v2.0 (`PEP 249`_), and also includes the :ref:`CrateDB dialect
+<using-sqlalchemy>` for `SQLAlchemy`_.
 
-    Check out the `sample application`_ (and the corresponding `sample
-    application documentation`_) for a practical demonstration of this driver
-    in use.
+The Python driver can be used to connect to both `CrateDB OSS`_ and `CrateDB
```
proddata, line 22: I don't think we refer to CrateDB as CrateDB OSS anywhere, do we?
Also I am not sure if we really should differentiate. CrateDB Cloud is mostly a CrateDB instance with user management and TLS enabled and therefore a "correct" setup CrateDB "OSS" instance would behave similar to a CrateDB Cloud instance

docs/index.rst
proddata: I think it doesn't really matter in this case. Double quotes in theory allow command substitution. But I think single quotes are more "correct" in this case.

The problematic part in `zsh` is using square brackets without quotes, as those are used for globbing

docs/index.rst
amotl: I also was not sure. I will remove that term again in order not to cause any confusion.

docs/sqlalchemy.rst
```diff
@@ -154,6 +154,19 @@ Once you have an CrateDB ``engine`` set up, you can create and use an SQLAlchemy
...
+.. _cloud-connect:
+
+Connecting to CrateDB Cloud
+...........................
+
+Connecting to `CrateDB Cloud`_ works like this. Please note the ``?ssl=true``
+query parameter at the end of the database URI.
+
+    >>> import sqlalchemy as sa
+    >>> engine = sa.create_engine("crate://admin:iM_n1%0xRzShsapU4U(lDGOj@example.aks1.westeurope.azure.cratedb.net:4200?ssl=true", echo=True)
```
proddata, line 167:
```suggestion
    >>> engine = sa.create_engine("crate://admin:pwd@example.aks1.westeurope.azure.cratedb.net:4200?ssl=true", echo=True)
```

should make it a bit easier to read?

amotl: Thanks. Maybe I will use `<PASSWORD>`, as we do in the Cloud documentation?

```python
client.connect("https://example.aks1.westeurope.azure.cratedb.net:4200", username="admin", password="<PASSWORD>", verify_ssl_cert=True)
```

docs/index.rst
amotl: Thanks. Fixed with 8235a75be0.

docs/index.rst
amotl: Thanks. Fixed with 8235a75be0.

docs/index.rst
amotl: Fixed with 8235a75be0.

docs/sqlalchemy.rst
amotl: Fixed with 14028867e.

proddata (approved):
great 👍

hammerhead (approved):

Issue #537, opened 2023-03-14
SQLAlchemy: Modernize batched inserts


amotl:
### Problem

There is an indication that the SQLAlchemy CrateDB dialect currently implements the [bulk_save_objects] method for bulk inserts, see [bulk_test.py#L70](https://github.com/crate/crate-python/blob/0.30.1/src/crate/client/sqlalchemy/tests/bulk_test.py#L70). The SQLAlchemy documentation says that this is a legacy method:

> This method is a legacy feature as of the 2.0 series of SQLAlchemy. For modern bulk INSERT and UPDATE, see the sections [ORM Bulk INSERT Statements](https://docs.sqlalchemy.org/en/20/orm/queryguide/dml.html#orm-queryguide-bulk-insert) and [ORM Bulk UPDATE by Primary Key](https://docs.sqlalchemy.org/en/20/orm/queryguide/dml.html#orm-queryguide-bulk-update).

### Analysis

[sqlalchemy_efficient_inserts.py](https://gist.github.com/amotl/d0fa29ac52d4c10e4a2544cffd606dfc) exercises two different bulk transfer options with SQLite, PostgreSQL and CrateDB. The outcome is that the PostgreSQL dialect will happily chunk the inserted records, correctly respecting the `insertmanyvalues_page_size` option. However, the CrateDB dialect does not do that, so that inserting >750k records will blow up the server with OOM errors.

```python
INSERT_RECORDS = 750_000
PAGE_SIZE = 10_000

records = [{"id": i, "name": f"foo_{i}"} for i in range(1, INSERT_RECORDS + 1)]
engine = sa.create_engine(dburi, insertmanyvalues_page_size=PAGE_SIZE, echo=True)

insertable = table.insert()
with engine.begin() as conn:
    conn.execute(insertable, parameters=records)
```

[bulk_save_objects]: https://docs.sqlalchemy.org/orm/session_api.html#sqlalchemy.orm.Session.bulk_save_objects

amotl:
With the patch from GH-539, this scenario will start working flawlessly, even with higher numbers in [`sqlalchemy_efficient_inserts.py`](https://gist.github.com/amotl/d0fa29ac52d4c10e4a2544cffd606dfc). That's how it should be.

```python
INSERT_RECORDS = 2_750_000
BATCHED_PAGE_SIZE = 20_000
```

amotl:
### SQLAlchemy Core
[...]

Issue #534, opened 2023-03-13
SQLAlchemy: `CompileError: The 'crate' dialect with current database version settings does not support in-place multirow inserts.`


amotl:
Hi,

when [evaluating loading Dask DataFrames into CrateDB](https://gist.github.com/amotl/c8039291fb4b4b16879fd8814de9ecdf) using the `method="multi"` option, the program croaks.

```python
Traceback (most recent call last):
  File "/path/to/lib/python3.10/site-packages/pandas/io/sql.py", line 1325, in insert_records
    return table.insert(chunksize=chunksize, method=method)
  File "/path/to/lib/python3.10/site-packages/pandas/io/sql.py", line 946, in insert
    num_inserted = exec_insert(conn, keys, chunk_iter)
  File "/path/to/lib/python3.10/site-packages/pandas/io/sql.py", line 869, in _execute_insert_multi
    result = conn.execute(stmt)
  File "/path/to/lib/python3.10/site-packages/sqlalchemy/engine/base.py", line 1414, in execute
    return meth(
  File "/path/to/lib/python3.10/site-packages/sqlalchemy/sql/elements.py", line 486, in _execute_on_connection
    return connection._execute_clauseelement(
  File "/path/to/lib/python3.10/site-packages/sqlalchemy/engine/base.py", line 1630, in _execute_clauseelement
    compiled_sql, extracted_params, cache_hit = elem._compile_w_cache(
  File "/path/to/lib/python3.10/site-packages/sqlalchemy/sql/elements.py", line 665, in _compile_w_cache
    compiled_sql = self._compiler(
  File "/path/to/lib/python3.10/site-packages/sqlalchemy/sql/elements.py", line 291, in _compiler
    return dialect.statement_compiler(dialect, self, **kw)
  File "/path/to/lib/python3.10/site-packages/sqlalchemy/sql/compiler.py", line 1269, in __init__
    Compiled.__init__(self, dialect, statement, **kwargs)
  File "/path/to/lib/python3.10/site-packages/sqlalchemy/sql/compiler.py", line 710, in __init__
    self.string = self.process(self.statement, **compile_kwargs)
  File "/path/to/lib/python3.10/site-packages/sqlalchemy/sql/compiler.py", line 755, in process
    return obj._compiler_dispatch(self, **kwargs)
  File "/path/to/lib/python3.10/site-packages/sqlalchemy/sql/visitors.py", line 143, in _compiler_dispatch
[...]

PR #538, opened 2023-03-14
SQLAlchemy: Support `INSERT...VALUES` with multiple value sets

amotl:
Enable `supports_multivalues_insert` on the CrateDB dialect, mitigating the error reported at GH-534:

    CompileError: The 'crate' dialect with current database version
    settings does not support in-place multirow inserts

> The Insert construct also supports being passed a list of dictionaries
> or full-table-tuples, which on the server will render the less common
> SQL syntax of "multiple values" - this syntax is supported on backends
> such as SQLite, PostgreSQL, MySQL, but not necessarily others.

> It is essential to note that passing multiple values is NOT the same
> as using traditional `executemany()` form. The above syntax is a special
> syntax not typically used. To emit an INSERT statement against
> multiple rows, the normal method is to pass a multiple values list to
> the `Connection.execute()` method [^1], which is supported by all database
> backends and is generally more efficient for a very large number of
> parameters [^2].

-- https://docs.sqlalchemy.org/core/dml.html#sqlalchemy.sql.expression.Insert.values.params.*args

[^1]: [Introduction to the traditional Core method of multiple parameter set invocation for INSERTs and other statements](https://docs.sqlalchemy.org/tutorial/dbapi_transactions.html#tutorial-multiple-parameters)
[^2]: With this method, there is curently an issue reported at GH-537.

/cc @SStorm, @WalBeh, @seut

### DIFF
--- a/src/crate/client/sqlalchemy/tests/compiler_test.py
+++ b/src/crate/client/sqlalchemy/tests/compiler_test.py
@@ -97,3 +97,33 @@ class SqlAlchemyCompilerTest(TestCase):
         selectable = self.mytable.select().offset(5).limit(42)
         statement = str(selectable.compile(bind=self.crate_engine))
         self.assertEqual(statement, "SELECT mytable.name, mytable.data \nFROM mytable \n LIMIT ? OFFSET ?")
+
+    def test_insert_multivalues(self):
+        """
+        Verify that "in-place multirow inserts" aka. "multivalues inserts" aka.
+        the `supports_multivalues_insert` dialect feature works.
+
+        When this feature is not enabled, using it will raise an error:
+
+            CompileError: The 'crate' dialect with current database version
+            settings does not support in-place multirow inserts
+
+        > The Insert construct also supports being passed a list of dictionaries
+        > or full-table-tuples, which on the server will render the less common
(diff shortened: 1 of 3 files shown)


src/crate/client/sqlalchemy/tests/compiler_test.py
```diff
@@ -97,3 +97,33 @@ def test_select_with_offset_and_limit(self):
...
+        > SQL syntax of "multiple values" - this syntax is supported on backends
+        > such as SQLite, PostgreSQL, MySQL, but not necessarily others.
+
+        > It is essential to note that passing multiple values is NOT the same
+        > as using traditional `executemany()` form. The above syntax is a special
+        > syntax not typically used. To emit an INSERT statement against
+        > multiple rows, the normal method is to pass a multiple values list to
+        > the `Connection.execute()` method, which is supported by all database
+        > backends and is generally more efficient for a very large number of
+        > parameters.
```
mfussenegger, line 122: Is there a reason it's not using `executemany` ? That would be more efficient for CrateDB - although the difference is probably small.

If not too much effort, maybe worth changing. Otherwise fine as is as well.

mfussenegger (approved):

PR #539, opened 2023-03-15
SQLAlchemy: Enable the `insertmanyvalues` feature for batch size control

amotl:
### About
SQLAlchemy's `insertmanyvalues` feature lets you control the batch size of `INSERT` operations using the `insertmanyvalues_page_size` option, which is applicable to all of the engine-, connection-, and statement- objects.

-- https://docs.sqlalchemy.org/core/connections.html#controlling-the-batch-size

### References
The patch will resolve this issue.
- GH-537

/cc @SStorm, @WalBeh, @seut, @hlcianfagna

### DIFF
--- a/CHANGES.txt
+++ b/CHANGES.txt
@@ -5,13 +5,19 @@ Changes for crate
 Unreleased
 ==========
 
-- SQLAlchemy: Support ``INSERT...VALUES`` with multiple value sets by enabling
+- SQLAlchemy Core: Support ``INSERT...VALUES`` with multiple value sets by enabling
   ``supports_multivalues_insert`` on the CrateDB dialect, it is used by pandas'
   ``method="multi"`` option
 
-- SQLAlchemy: Enable the ``insertmanyvalues`` feature, which lets you control
+- SQLAlchemy Core: Enable the ``insertmanyvalues`` feature, which lets you control
   the batch size of ``INSERT`` operations using the ``insertmanyvalues_page_size``
   engine-, connection-, and statement-options.
+
+- SQLAlchemy ORM: Remove support for the legacy ``session.bulk_save_objects`` API
+  on SQLAlchemy 2.0, in favor of the new ``insertmanyvalues`` feature. Performance
+  optimizations from ``bulk_save()`` have been made inherently part of ``add_all()``.
+  Note: The legacy mode will still work on SQLAlchemy 1.x, while SQLAlchemy 2.x users
+  MUST switch to the new method now.
 
 
 2023/03/02 0.30.1
--- a/CHANGES.txt
+++ b/CHANGES.txt
@@ -9,6 +9,10 @@ Unreleased
   ``supports_multivalues_insert`` on the CrateDB dialect, it is used by pandas'
   ``method="multi"`` option
 
+- SQLAlchemy: Enable the ``insertmanyvalues`` feature, which lets you control
+  the batch size of ``INSERT`` operations using the ``insertmanyvalues_page_size``
+  engine-, connection-, and statement-options.
+
 
 2023/03/02 0.30.1
 =================
--- a/src/crate/client/sqlalchemy/tests/bulk_test.py
+++ b/src/crate/client/sqlalchemy/tests/bulk_test.py
@@ -19,11 +19,14 @@
 # with Crate these terms will supersede the license and you may use the
 # software solely pursuant to the terms of the relevant commercial agreement.
 
-from unittest import TestCase
+from unittest import TestCase, skipIf
 from unittest.mock import patch, MagicMock
 
 import sqlalchemy as sa
 from sqlalchemy.orm import Session
+
+from crate.client.sqlalchemy.sa_version import SA_VERSION, SA_2_0
+
 try:
     from sqlalchemy.orm import declarative_base
 except ImportError:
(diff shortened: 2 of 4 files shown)


mfussenegger (approved):

matriv (approved):
nice! thank you!

amotl:
This patch apparently breaks other tests. I will have to investigate.

amotl:
> This patch apparently breaks other tests.

Indeed. Enabling `use_insertmanyvalues_wo_returning = True` makes the `SqlAlchemyBulkTest.test_bulk_save` fail. Because this area how bulk inserts may have been handled before, seems to be declared as "legacy" now, it may make sense to remove it altogether.

#### References

- https://docs.sqlalchemy.org/orm/queryguide/dml.html#legacy-session-bulk-insert-methods
- https://docs.sqlalchemy.org/orm/queryguide/dml.html#orm-queryguide-legacy-bulk-update
- https://docs.sqlalchemy.org/en/20/orm/session_api.html#sqlalchemy.orm.Session.bulk_save_objects

amotl:
### Recap / Introduction
> Enabling `use_insertmanyvalues_wo_returning = True` makes the `SqlAlchemyBulkTest.test_bulk_save` fail.

> **Problem:** There is an indication that the SQLAlchemy CrateDB dialect currently only implements the [bulk_save_objects] method for bulk inserts [...]. This method is a legacy feature as of the 2.0 series of SQLAlchemy [...].
> - GH-537

### Solution
8073178cc provides a modernized version of this test case for SA20, using `session.add_all()` instead of the legacy `session.bulk_save_objects()`. The rationale is provided in the "Details" section.

### Details

Verify modern SQLAlchemy bulk INSERT mode.

> A list of parameter dictionaries sent to the `Session.execute.params` parameter,
> separate from the Insert object itself, will invoke *bulk INSERT mode* for the
> statement, which essentially means the operation will optimize as much as
> possible for many rows.
>
> -- https://docs.sqlalchemy.org/orm/queryguide/dml.html#orm-queryguide-bulk-insert

> We have been looking into getting performance optimizations
> from `bulk_save()` to be inherently part of `add_all()`.
>
> -- https://github.com/sqlalchemy/sqlalchemy/discussions/6935#discussioncomment-1233465

> The remaining performance limitation, that the `cursor.executemany()` DBAPI method
> does not allow for rows to be fetched, is resolved for most backends by *foregoing*
> the use of `executemany()` and instead restructuring individual INSERT statements
> to each accommodate a large number of rows in a single statement that is invoked
> using `cursor.execute()`. This approach originates from the `psycopg2` fast execution
> helpers feature of the `psycopg2` DBAPI, which SQLAlchemy incrementally added more
> and more support towards in recent release series.
>
> -- https://docs.sqlalchemy.org/core/connections.html#engine-insertmanyvalues


### Differences
For your information, this diff outlines the differences between the legacy and the modern variant. Because the `test_bulk_save_modern` test case method has been _added_, this would otherwise not be visible when looking at the patch.
```diff
@@ -1,9 +1,13 @@
-@skipIf(SA_VERSION >= SA_2_0, "SQLAlchemy 2.0 uses modern bulk INSERT mode")
+@skipIf(SA_VERSION <= SA_1_4, "SQLAlchemy 1.x uses legacy bulk INSERT mode")
 @patch('crate.client.connection.Cursor', FakeCursor)
-def test_bulk_save_legacy(self):
+def test_bulk_save_modern(self):
     """
-    Verify legacy SQLAlchemy bulk INSERT mode.
+    Verify modern SQLAlchemy bulk INSERT mode.
     """
+
+    # Adjust diff output on errors.
+    self.maxDiff = None
+
     chars = [
         self.character(name='Arthur', age=35),
         self.character(name='Banshee', age=26),
@@ -12,20 +16,21 @@

     fake_cursor.description = ()
     fake_cursor.rowcount = len(chars)
-    fake_cursor.executemany.return_value = [
+    fake_cursor.execute.return_value = [
         {'rowcount': 1},
         {'rowcount': 1},
         {'rowcount': 1},
     ]
-    self.session.bulk_save_objects(chars)
-    (stmt, bulk_args), _ = fake_cursor.executemany.call_args
+    self.session.add_all(chars)
+    self.session.commit()
+    (stmt, bulk_args), _ = fake_cursor.execute.call_args

-    expected_stmt = "INSERT INTO characters (name, age) VALUES (?, ?)"
+    expected_stmt = "INSERT INTO characters (name, age) VALUES (?, ?), (?, ?), (?, ?)"
     self.assertEqual(expected_stmt, stmt)

     expected_bulk_args = (
-        ('Arthur', 35),
-        ('Banshee', 26),
-        ('Callisto', 37)
+        'Arthur', 35,
+        'Banshee', 26,
+        'Callisto', 37,
     )
     self.assertSequenceEqual(expected_bulk_args, bulk_args)
```

[bulk_save_objects]: https://docs.sqlalchemy.org/orm/session_api.html#sqlalchemy.orm.Session.bulk_save_objects

amotl:
@mfussenegger and @matriv: I needed to add more changes to satisfy the `SqlAlchemyBulkTest`. Can I ask for another review? Thank you already!

CHANGES.txt
```diff
@@ -5,10 +5,20 @@ Changes for crate
...
-- SQLAlchemy: Support ``INSERT...VALUES`` with multiple value sets by enabling
+- SQLAlchemy Core: Support ``INSERT...VALUES`` with multiple value sets by enabling
   ``supports_multivalues_insert`` on the CrateDB dialect, it is used by pandas'
   ``method="multi"`` option
 
+- SQLAlchemy Core: Enable the ``insertmanyvalues`` feature, which lets you control
+  the batch size of ``INSERT`` operations using the ``insertmanyvalues_page_size``
+  engine-, connection-, and statement-options.
+
+- SQLAlchemy ORM: Remove support for the legacy ``session.bulk_save_objects`` API
```
matriv, line 16: Shouldn't we mark this a breaking change? (we can also do during release of course)

src/crate/client/sqlalchemy/tests/bulk_test.py
```diff
@@ -79,3 +109,62 @@ def test_bulk_save(self):
...
+        > to each accommodate a large number of rows in a single statement that is invoked
+        > using `cursor.execute()`. This approach originates from the `psycopg2` fast execution
+        > helpers feature of the `psycopg2` DBAPI, which SQLAlchemy incrementally added more
+        > and more support towards in recent release series.
+        >
+        > -- https://docs.sqlalchemy.org/core/connections.html#engine-insertmanyvalues
+        """
+
+        # Adjust diff output on errors.
+        self.maxDiff = None
```
matriv, line 143: What's the purpose of this, I don't get it, sorry.

matriv (approved):
LGTM. Thx a lot for the detailed explanation @amotl !!
Left a couple of questions.

src/crate/client/sqlalchemy/tests/bulk_test.py
amotl: Thanks. I've amended the patch and rephrased this comment. Is it better?
```python
# Don't truncate unittest's diff output on `assertListEqual`.
self.maxDiff = None
```

CHANGES.txt
amotl: Thanks. Initially, I had a strong BREAKING CHANGE admonition here. But after observing that the legacy bulk API will still work for users on SA 1.x, and given that support for SA 2.x has only been added recently, I relaxed that note to:
```
Note: The legacy mode will still work on SQLAlchemy 1.x, while SQLAlchemy 2.x users
MUST switch to the new method now.
```
Please let me know if you think we should bring back the stronger note.

amotl:
Thanks for the review, @matriv. I hope your comments have been addressed well. I will let it sitting until @mfussenegger can have another look.

PR #552, opened 2023-05-10
Documentation: Add executable code examples

amotl:
Hi there,

instead of only nurturing the [cratedb-examples » by-language] folder with executable and sensible code examples, which demonstrate and explore specific areas around communicating with CrateDB efficiently, I think it is also valuable to maintain corresponding `examples` folders within the repositories itself, because this is where developers very often will look for, just because, well, runnable code examples are just so incredibly useful, and where else should I expect to find them? [^1]

In this spirit, this patch is adding two code example from my gists collection, [sqlalchemy_efficient_inserts.py] and [cratedb_pandas_insert.py], which I think make sense to be added here. I don't think they _necessarily_ have to be tested automatically, but sure it will be nice to have, so if you think they should be tested, I am sure we will find a way to add them to the test suite without too much obstruction on a subsequent iteration.

With kind regards,
Andreas.

P.S.: This patch is here to set the stage for a subsequent one, which refactors the performance-optimized `insert_bulk` function into the package itself, in order to be able to reuse it in downstream software, notebooks and tutorials more easily.

[cratedb-examples » by-language]: https://github.com/crate/cratedb-examples/tree/main/by-language
[sqlalchemy_efficient_inserts.py]: https://gist.github.com/amotl/d0fa29ac52d4c10e4a2544cffd606dfc
[cratedb_pandas_insert.py]: https://gist.github.com/amotl/d39db8eff6ad5de5898ce61fac4c0af3

[^1]: I found this has not been much of a guideline at CrateDB's auxilliary repositories yet, so I started that recently at [CrateDB PDO driver » Examples](https://github.com/crate/crate-pdo/tree/main/examples).

/cc @marijaselakovic, @hammerhead, @hlcianfagna, @proddata, @WalBeh

### DIFF
--- a/.readthedocs.yml
+++ b/.readthedocs.yml
@@ -1,0 +1,25 @@
+# .readthedocs.yml
+# Read the Docs configuration file
+
+# Details
+# - https://docs.readthedocs.io/en/stable/config-file/v2.html
+
+# Required
+version: 2
+
+build:
+  os: "ubuntu-22.04"
+  tools:
+    python: "3.11"
+
+# Build documentation in the docs/ directory with Sphinx
+sphinx:
+  configuration: docs/conf.py
+
+python:
+  install:
+    - requirements: docs/requirements.txt
+
+# Optionally build your docs in additional formats such as PDF
+# formats:
+#   - pdf
--- a/examples/insert_efficient.py
+++ b/examples/insert_efficient.py
@@ -1,0 +1,138 @@
+"""
+About
+=====
+
+Example program to demonstrate multi-row inserts and batched inserts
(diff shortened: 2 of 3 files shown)


matriv:
In my opinion we shouldn't duplicate work to each repo, I believe it's enough to have all examples in the `cratedb-examples` repo and link to them?

amotl:
Hi @matriv,

your suggestion resonates well with me. I will adjust the patches correspondingly.

With kind regards,
Andreas.

amotl:
This patch is being converged into https://github.com/crate/cratedb-examples/pull/64. Thanks.

Issue #425, opened 2022-06-01
SQLAlchemy 1.4 aftermath


amotl:
Hi there,

while working on #391, some backlog items have accumulated. I will gather them within this ticket.

- [ ] Migrate from `before_execute` event to `ExecutionContext.pre_exec()`
      Discussion: https://github.com/sqlalchemy/sqlalchemy/discussions/5915#discussioncomment-2782664

- [ ] Investigate `CompileError: Unconsumed column names`
      Discussion: https://github.com/crate/crate-python/pull/391#pullrequestreview-989026539

- [ ] https://github.com/crate/crate-python/issues/423

With kind regards,
Andreas.

robd003:
Getting support for async SQLAlchemy would be super useful

I have a few queries that can take slightly over 1 sec to execute and being able to not block would be **HUGE**

amotl:
Dear Robert,

support for asynchronous communication with SQLAlchemy, based on the asyncpg and psycopg3 drivers, is being evaluated at GH-532. Please note that this is experimental, and we currently have no schedule about when or how this will be released.

With kind regards,
Andreas.

amotl:
Coming from https://github.com/crate/crate-python/pull/553#issuecomment-1545704082, there are two additional backlog items:

- [ ] Expand support for fast-path bulk insert operations to the other DML operations `UPDATE` and `DELETE` offered by the CrateDB bulk operations endpoint, when this is sensible within a pandas/Dask environment.
- [ ] Don't stop at `to_sql()`, also cover optimal `read_sql()` techniques within the documentation, when possible.

PR #553, opened 2023-05-11
Add fast-path INSERT method `insert_bulk` for SQLAlchemy/pandas/Dask

amotl:
Hi again,

in the same spirit as https://github.com/crate/crate-pdo/pull/143, this patch unlocks [CrateDB bulk operations] for efficient batch inserts using [pandas](https://pandas.pydata.org/) and [Dask](https://www.dask.org/). It will accompany a followup to the [Data Processing and Analytics with Dask and CrateDB: A Step-by-Step Tutorial].

The corresponding documentation section can be inspected at [Preview: SQLAlchemy: DataFrame operations](https://crate-python--553.org.readthedocs.build/en/553/by-example/sqlalchemy/dataframe.html).

With kind regards,
Andreas.

[CrateDB bulk operations]: https://crate.io/docs/crate/reference/en/latest/interfaces/http.html#bulk-operations
[Data Processing and Analytics with Dask and CrateDB: A Step-by-Step Tutorial]: https://community.crate.io/t/data-processing-and-analytics-with-dask-and-cratedb-a-step-by-step-tutorial/1482

/cc @marijaselakovic, @hammerhead, @hlcianfagna, @proddata, @WalBeh

### DIFF
--- a/docs/by-example/sqlalchemy/dataframe.rst
+++ b/docs/by-example/sqlalchemy/dataframe.rst
@@ -5,6 +5,12 @@
 SQLAlchemy: DataFrame operations
 ================================
 
+.. rubric:: Table of Contents
+
+.. contents::
+   :local:
+
+
 About
 =====
 
@@ -12,7 +18,7 @@ This section of the documentation demonstrates support for efficient batch/bulk
 ``INSERT`` operations with `pandas`_ and `Dask`_, using the CrateDB SQLAlchemy dialect.
 
 Efficient bulk operations are needed for typical `ETL`_ batch processing and
-data streaming workloads, for example to move data in- and out of OLAP data
+data streaming workloads, for example to move data in and out of OLAP data
 warehouses, as contrasted to interactive online transaction processing (OLTP)
 applications. The strategies of `batching`_ together series of records for
 improving performance are also referred to as `chunking`_.
@@ -21,6 +27,8 @@ improving performance are also referred to as `chunking`_.
 Introduction
 ============
 
+pandas
+------
 The :ref:`pandas DataFrame <pandas:api.dataframe>` is a structure that contains
 two-dimensional data and its corresponding labels. DataFrames are widely used
 in data science, machine learning, scientific computing, and many other
@@ -34,11 +42,29 @@ powerful than tables or spreadsheets because they are an integral part of the
 The :ref:`pandas I/O subsystem <pandas:api.io>` for `relational databases`_
 using `SQL`_ is based on `SQLAlchemy`_.
 
+Dask
+----
+`Dask`_ is a flexible library for parallel computing in Python, which scales
+Python code from multi-core local machines to large distributed clusters in
+the cloud. Dask provides a familiar user interface by mirroring the APIs of
+other libraries in the PyData ecosystem, including `pandas`_, `scikit-learn`_,
+and `NumPy`_.
 
-.. rubric:: Table of Contents
+A :doc:`dask:dataframe` is a large parallel DataFrame composed of many smaller
+pandas DataFrames, split along the index. These pandas DataFrames may live on
+disk for larger-than-memory computing on a single machine, or on many different
+machines in a cluster. One Dask DataFrame operation triggers many operations on
+the constituent pandas DataFrames.
 
-.. contents::
-   :local:
+
+Compatibility notes
+===================
+
+.. NOTE::
+
+    Please note that DataFrame support for pandas and Dask is only validated
+    with Python 3.8 and higher, and SQLAlchemy 1.4 and higher. We recommend
+    to use the most recent versions of those libraries.
 
 
 Efficient ``INSERT`` operations with pandas
@@ -118,6 +144,91 @@ workload across multiple batches, using a defined chunk size.
     tutorial <wide-narrow-pandas-tutorial_>`_ about the same topic.
 
 
+Efficient ``INSERT`` operations with Dask
+=========================================
+
+The same ``bulk_insert`` function presented in the previous section will also
+be used in the context of `Dask`_, in order to make the
+:func:`dask:dask.dataframe.to_sql` method more efficiently, based on the
+`CrateDB bulk operations`_ endpoint.
+
+The example below will partition your insert workload into equal-sized parts, and
+schedule it to be executed on Dask cluster resources, using a defined number of
+compute partitions. Each worker instance will then insert its partition's records
+in a batched/chunked manner, using a defined chunk size, effectively using the
+pandas implementation introduced in the previous section.
+
+    >>> import dask.dataframe as dd
+    >>> from pandas._testing import makeTimeDataFrame
+    >>> from crate.client.sqlalchemy.support import insert_bulk
+    ...
+    >>> # Define the number of records, the number of computing partitions,
+    >>> # and the chunk size of each database insert operation.
+    >>> INSERT_RECORDS = 100
+    >>> NPARTITIONS = 4
+    >>> CHUNK_SIZE = 25
+    ...
+    >>> # Create a Dask DataFrame.
+    >>> df = makeTimeDataFrame(nper=INSERT_RECORDS, freq="S")
+    >>> ddf = dd.from_pandas(df, npartitions=NPARTITIONS)
+    ...
+    >>> # Insert content of DataFrame using multiple workers on a
+    >>> # compute cluster, transferred using batches of records.
+    >>> ddf.to_sql(
+    ...     name="test-testdrive",
+    ...     uri=f"crate://{crate_host}",
+    ...     if_exists="replace",
+    ...     index=False,
+    ...     chunksize=CHUNK_SIZE,
+    ...     method=insert_bulk,
+    ...     parallel=True,
+    ... )
+
+
+.. TIP::
+
+    You will observe that optimizing your workload will now also involve determining a
+    good value for the ``NPARTITIONS`` argument, based on the capacity and topology of
+    the available compute resources, and based on workload characteristics or policies
+    like peak- vs. balanced- vs. shared-usage. For example, on a machine or cluster fully
+    dedicated to the problem at hand, you may want to use all available processor cores,
+    while on a shared system, this strategy may not be appropriate.
+
+    If you want to dedicate all available compute resources on your machine, you may want
+    to use the number of CPU cores as a value to the ``NPARTITIONS`` argument. You can find
+    out about the available CPU cores on your machine, for example by running the ``nproc``
+    command in your terminal.
+
+    Depending on the implementation and runtime behavior of the compute task, the optimal
+    number of worker processes, determined by the ``NPARTITIONS`` argument, also needs to be
+    figured out by running a few test iterations. For that purpose, you can use the
+    `insert_dask.py`_ program as a blueprint.
+
+    Adjusting this value in both directions is perfectly fine: If you observe that you are
+    overloading the machine, maybe because there are workloads scheduled other than the one
+    you are running, try to reduce the value. If fragments/steps of your implementation
+    involve waiting for network or disk I/O, you may want to increase the number of workers
+    beyond the number of available CPU cores, to increase utilization. On the other hand,
+    you should be wary about not over-committing resources too much, as it may slow your
+    system down.
+
+    Before getting more serious with Dask, you are welcome to read and watch the excellent
+    :doc:`dask:best-practices` and :ref:`dask:dataframe.performance` resources, in order to
+    learn about things to avoid, and beyond. For finding out if your compute workload
+    scheduling is healthy, you can, for example, use Dask's :doc:`dask:dashboard`.
+
+.. WARNING::
+
+    Because the settings assigned in the example above fit together well, the ``to_sql()``
+    instruction will effectively run four insert operations, executed in parallel, and
+    scheduled optimally on the available cluster resources.
+
+    However, not using those settings sensibly, you can easily misconfigure the resource
+    scheduling system, and overload the underlying hardware or operating system, virtualized
+    or not. This is why experimenting with different parameters, and a real dataset, is crucial.
+
+
+
 .. hidden: Disconnect from database
 
     >>> engine.dispose()
@@ -126,7 +237,10 @@ workload across multiple batches, using a defined chunk size.
 .. _batching: https://en.wikipedia.org/wiki/Batch_processing#Common_batch_processing_usage
 .. _chunking: https://en.wikipedia.org/wiki/Chunking_(computing)
 .. _CrateDB bulk operations: https://crate.io/docs/crate/reference/en/latest/interfaces/http.html#bulk-operations
+.. _Dask: https://en.wikipedia.org/wiki/Dask_(software)
 .. _DataFrame computing: https://realpython.com/pandas-dataframe/
+.. _ETL: https://en.wikipedia.org/wiki/Extract,_transform,_load
+.. _insert_dask.py: https://github.com/crate/cratedb-examples/blob/main/by-language/python-sqlalchemy/insert_dask.py
 .. _insert_pandas.py: https://github.com/crate/cratedb-examples/blob/main/by-language/python-sqlalchemy/insert_pandas.py
 .. _leveling up to 200_000: https://acepor.github.io/2017/08/03/using-chunksize/
 .. _NumPy: https://en.wikipedia.org/wiki/NumPy
@@ -134,6 +248,8 @@ workload across multiple batches, using a defined chunk size.
 .. _pandas DataFrame: https://pandas.pydata.org/pandas-docs/stable/reference/frame.html
 .. _Python: https://en.wikipedia.org/wiki/Python_(programming_language)
 .. _relational databases: https://en.wikipedia.org/wiki/Relational_database
+.. _scikit-learn: https://en.wikipedia.org/wiki/Scikit-learn
+.. _SNAT port exhaustion: https://learn.microsoft.com/en-us/azure/load-balancer/troubleshoot-outbound-connection
 .. _SQL: https://en.wikipedia.org/wiki/SQL
 .. _SQLAlchemy: https://aosabook.org/en/v2/sqlalchemy.html
 .. _the chunksize should not be too small: https://acepor.github.io/2017/08/03/using-chunksize/
--- a/docs/by-example/sqlalchemy/dataframe.rst
+++ b/docs/by-example/sqlalchemy/dataframe.rst
@@ -1,0 +1,138 @@
+.. _sqlalchemy-pandas:
+.. _sqlalchemy-dataframe:
+
+================================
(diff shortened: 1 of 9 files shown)


docs/by-example/sqlalchemy/dataframe.rst
```diff
@@ -0,0 +1,106 @@
...
+    It is a good idea to start your explorations with a chunk size of 5000, and
+    then see if performance improves when you increase or decrease that figure.
+    Chunk sizes of 20000 may also be applicable, but make sure to take the limits
+    of your HTTP infrastructure into consideration.
+
+    In order to learn more about what wide- vs. long-form (tidy, stacked, narrow)
+    data means in the context of `DataFrame computing`_, let us refer you to `a
+    general introduction <wide-narrow-general_>`_, the corresponding section in
+    the `Data Computing book <wide-narrow-data-computing_>`_, and a `pandas
+    tutorial <wide-narrow-pandas-tutorial_>`_ about the same topic.
```
amotl, line 86: On this section, about choosing the optimal batch size, @marijaselakovic had a good suggestion:

> Looks good, I am just thinking maybe it makes sense to add that when choosing “optimal” chunksize to consider besides the share and size of the data, also the memory of the machine. I find [here](https://docs.dask.org/en/stable/array-chunks.html) that chunk should be small enough to fit the memory but large enough to minimize the overhead of data ingestion. Even though this is from Dask documentation, should apply to pandas too. The rest looks good!

Thank you!

amotl: @all: Please let me know if you can think of any other details I might have missed in this context of optimally educating users about this topic. Now, it's the right time to bring them in. Thank you already!

amotl: Hi again. I've improved the documentation section with 837426bff4e. I did not use the link to Dask yet, because the section currently only explains efficient batching on behalf of pandas, and I do not want to confuse users too much. Another section about Dask should probably be added right away, in order to better outline the _parallel_ computing aspects around it.

amotl:
Regarding the failing software tests, I will exclude the relevant test matrix slots from being executed, and add a corresponding `.. admonition` section to the documentation, about which versions of Python and SQLAlchemy are supported.

docs/by-example/sqlalchemy/dataframe.rst
```diff
@@ -0,0 +1,121 @@
...
+   :local:
+
+
+Efficient ``INSERT`` operations with pandas
+===========================================
+
+The package provides a ``bulk_insert`` function to use the
+:meth:`pandas:pandas.DataFrame.to_sql` method most efficiently, based on the `CrateDB
+bulk operations`_ endpoint. It will effectively split your insert workload across
+multiple batches, using a defined chunk size.
```
hlcianfagna, line 44:
```suggestion
:meth:`pandas:pandas.DataFrame.to_sql` method more efficiently, based on the
`CrateDB bulk operations`_ endpoint. It will effectively split your insert
workload across multiple batches, using a defined chunk size.
```

docs/by-example/sqlalchemy/dataframe.rst
```diff
@@ -0,0 +1,121 @@
...
+
+    You will need to determine a good chunk size by running corresponding experiments
+    on your own behalf. For that purpose, you can use the `insert_pandas.py`_ program
+    as a blueprint.
+
+    It is a good idea to start your explorations with a chunk size of 5_000, and
+    then see if performance improves when you increase or decrease that figure.
+    People are reporting that 10_000-20_000 is their optimal setting, but if you
+    process, for example, just three "small" columns, you may also experiment with
+    `leveling up to 200_000`_, because `the chunksize should not be too small`_.
```
hlcianfagna, line 91: Any reason for using `_` instead of `,` as thousands separator?

docs/by-example/sqlalchemy/dataframe.rst
```diff
@@ -0,0 +1,121 @@
...
+    your data, specifically the width of each record, i.e. the number of columns
+    and their individual sizes, which will in the end determine the total size of
+    each batch/chunk.
+
+    Two specific things should be taken into consideration when determining the
+    optimal chunk size for a specific dataset. First, when working with data
+    larger than main memory available on your machine, each chunk should be small
+    enough to fit into the memory, but large enough to minimize the overhead of a
+    single data insert operation. Second, as each batch is submitted using HTTP,
+    you should know about the request size limits of your HTTP infrastructure.
```
hlcianfagna, line 81: Maybe clarify what is meant by http infrastructure with an example, something like consider your infrastructure may include device like proxy servers or intrusion protection systems which may reject or slow down abnormally large requests.

src/crate/client/sqlalchemy/support.py
```diff
@@ -0,0 +1,62 @@
...
+
+    # For debugging and tracing the batches running through this method.
+    # Because it's a performance-optimized code path, the log statements are not active by default.
+    # logger.info(f"Bulk SQL:     {sql}")
+    # logger.info(f"Bulk records: {len(data)}")
+    # logger.info(f"Bulk data:    {data}")
+
+    # Invoke bulk insert operation.
+    cursor = conn._dbapi_connection.cursor()
+    cursor.execute(sql=sql, bulk_parameters=data)
```
hlcianfagna, line 61: Does this translate to `bulk_args` in the actual http request?

docs/by-example/sqlalchemy/dataframe.rst
amotl: It is valid Python syntax, in order to get around the ambiguity with metric vs. imperial systems, and I think it is legitimate to use it within corresponding Python documentation.

src/crate/client/sqlalchemy/support.py
amotl: Yes, that's correct.

docs/by-example/sqlalchemy/dataframe.rst
```diff
@@ -0,0 +1,121 @@
...
+    You will observe that the optimal chunk size highly depends on the shape of
+    your data, specifically the width of each record, i.e. the number of columns
+    and their individual sizes, which will in the end determine the total size of
+    each batch/chunk.
+
+    Two specific things should be taken into consideration when determining the
+    optimal chunk size for a specific dataset. First, when working with data
+    larger than main memory available on your machine, each chunk should be small
+    enough to fit into the memory, but large enough to minimize the overhead of a
+    single data insert operation. Second, as each batch is submitted using HTTP,
```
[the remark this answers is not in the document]
amotl: Thanks, I've tried to address it with 13f4b7b25f5.

docs/by-example/sqlalchemy/dataframe.rst
amotl: I tried to improve with 13f4b7b25f5, also on this matter. Thanks.

docs/by-example/sqlalchemy/dataframe.rst
```diff
@@ -0,0 +1,257 @@
...
+
+
+About
+=====
+
+This section of the documentation demonstrates support for efficient batch/bulk
+``INSERT`` operations with `pandas`_ and `Dask`_, using the CrateDB SQLAlchemy dialect.
+
+Efficient bulk operations are needed for typical `ETL`_ batch processing and
+data streaming workloads, for example to move data in- and out of OLAP data
```
seut, line 21:
```suggestion
data streaming workloads, for example to move data in and out of OLAP data
```

src/crate/client/sqlalchemy/support.py
```diff
@@ -0,0 +1,62 @@
...
+
+    # Compile SQL statement and materialize batch.
+    sql = str(pd_table.table.insert().compile(bind=conn))
+    data = list(data_iter)
+
+    # For debugging and tracing the batches running through this method.
+    # Because it's a performance-optimized code path, the log statements are not active by default.
+    # logger.info(f"Bulk SQL:     {sql}")
+    # logger.info(f"Bulk records: {len(data)}")
+    # logger.info(f"Bulk data:    {data}")
```
seut, line 57: Can't this be simplified by changing the level to `DEBUG` and surround it with something like `if logger.level is DEBUG` (don't know the exact syntax) to avoid any serialization?

src/crate/client/sqlalchemy/tests/bulk_test.py
```diff
@@ -168,3 +168,85 @@ def test_bulk_save_modern(self):
...
+        from crate.client.sqlalchemy.support import insert_bulk
+
+        # 42 records / 4 partitions means each partition has a size of 10.5 elements.
+        # Because the chunk size 8 is slightly smaller than 10, the partition will not
+        # fit into it, so two batches will be emitted to the database for each data
+        # partition. 4 partitions * 2 batches = 8 insert operations will be emitted.
+        INSERT_RECORDS = 42
+        NPARTITIONS = 4
+        CHUNK_SIZE = 8
+        OPCOUNT = math.ceil(INSERT_RECORDS / NPARTITIONS / CHUNK_SIZE) * NPARTITIONS
```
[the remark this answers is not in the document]
seut: Can you add the comment?

seut (commented):
Nice! added some suggestions

amotl:
Hi.

Thanks for all the excellent review comments.

> [This patch] will accompany a followup to the [Data Processing and Analytics with Dask and CrateDB: A Step-by-Step Tutorial] by @marijaselakovic.

Initially, I rushed a bit to bring this in in time for this tutorial, but I missed the deadline. Now, that we have a little more headroom, I think this patch should only be the foundation, and those aspects should be addresses by corresponding followup patches:

- Expand support for fast-path bulk insert operations to the other DML operations `UPDATE` and `DELETE` offered by the CrateDB bulk operations endpoint, when this is sensible within a pandas/Dask environment.
- Don't stop at `to_sql()`, also cover optimal `read_sql()` techniques within the documentation, when possible.

I hope you agree with that strategy.

With kind regards,
Andreas.

[Data Processing and Analytics with Dask and CrateDB: A Step-by-Step Tutorial]: https://community.crate.io/t/data-processing-and-analytics-with-dask-and-cratedb-a-step-by-step-tutorial/1482

hlcianfagna (approved):
Great work @amotl

docs/by-example/sqlalchemy/dataframe.rst
```diff
@@ -0,0 +1,257 @@
...
+    ...     method=insert_bulk,
+    ...     parallel=True,
+    ... )
+
+
+.. TIP::
+
+    You will observe that optimizing your workload will now also involve determining a
+    good value for the ``NPARTITIONS`` argument, based on the capacity and topology of
+    the available compute resources , and based on workload characteristics or policies
```
amotl, line 192:
```suggestion
    the available compute resources, and based on workload characteristics or policies
```

docs/by-example/sqlalchemy/dataframe.rst
```diff
@@ -0,0 +1,257 @@
...
+    like peak- vs. balanced- vs. shared-usage. For example, on a machine or cluster fully
+    dedicated to the problem at hand, you may want to use all available processor cores,
+    while on a shared system, this strategy may not be appropriate.
+
+    If you want to dedicate all available compute resources on your machine, you may want
+    to use the number of CPU cores as a value to the ``NPARTITIONS`` argument. You can find
+    out about the available CPU core on your machine, for example by running the ``nproc``
```
amotl, line 199:
```suggestion
    out about the available CPU cores on your machine, for example by running the ``nproc``
```

docs/by-example/sqlalchemy/dataframe.rst
```diff
@@ -0,0 +1,257 @@
...
+    ...
+    >>> # Define the number of records, the number of computing partitions,
+    >>> # and the chunk size of each database insert operation.
+    >>> INSERT_RECORDS = 100
+    >>> NPARTITIONS = 4
+    >>> CHUNK_SIZE = 25
+    ...
+    >>> # Create a Dask DataFrame.
+    >>> df = makeTimeDataFrame(nper=INSERT_RECORDS, freq="S")
+    >>> ddf = dd.from_pandas(df, npartitions=NPARTITIONS)
```
amotl, line 173: Hi again,

at https://github.com/crate/crate-python/pull/553#issuecomment-1545704082, I was enumerating potential subsequent iterations on this patch. Another detail could be to elaborate a bit about the other main argument to the `dask.from_pandas()` function, `chunksize=`, which can only be used _exclusively_ to the `npartitions=` argument.
```python
ddf = dask.from_pandas(df, chunksize=CHUNKSIZE)
```

In this way, you don't specify the number of partitions, but also use the **chunk size** as a parameter to configure the workload scheduler. This concept might fit even better with typical ETL tasks from/to database systems, which we are exploring here to make them more efficient with CrateDB.

You'd probably use the same `chunksize` value here, which will also be used for configuring the outbound batch chunker to the database, but I am not sure about it yet. If this is the case, it would make its usage even easier in different scenarios, because a user would just need to configure a good chunk size, not different to the basic pandas usage at all, and not need to spend much specific thoughts on compute resources at all.

With kind regards,
Andreas.

----

```python
def from_pandas(
    data: pd.DataFrame | pd.Series,
    npartitions: int | None = None,
    chunksize: int | None = None,
    sort: bool = True,
    name: str | None = None,
) -> DataFrame | Series:
    """
    Construct a Dask DataFrame from a Pandas DataFrame

    This splits an in-memory Pandas dataframe into several parts and constructs
    a dask.dataframe from those parts on which Dask.dataframe can operate in
    parallel.  By default, the input dataframe will be sorted by the index to
    produce cleanly-divided partitions (with known divisions).  To preserve the
    input ordering, make sure the input index is monotonically-increasing. The
    ``sort=False`` option will also avoid reordering, but will not result in
    known divisions.

    Note that, despite parallelism, Dask.dataframe may not always be faster
    than Pandas.  We recommend that you stay with Pandas for as long as
    possible before switching to Dask.dataframe.

    npartitions : int, optional
        The number of partitions of the index to create. Note that if there
        are duplicate values or insufficient elements in ``data.index``, the
        output may have fewer partitions than requested.
    chunksize : int, optional
        The desired number of rows per index partition to use. Note that
        depending on the size and index of the dataframe, actual partition
        sizes may vary.
    """
```

docs/by-example/sqlalchemy/dataframe.rst
```diff
@@ -0,0 +1,257 @@
...
+    number of worker processes, determined by the ``NPARTITIONS`` argument, also needs to be
+    figured out by running a few test iterations. For that purpose, you can use the
+    `insert_dask.py`_ program as a blueprint.
+
+    Adjusting this value in both directions is perfectly fine: If you observe that you are
+    overloading the machine, maybe because there are workloads scheduled other than the one
+    you are running, try to reduce the value. If fragments/steps of your implementation
+    involve waiting for network or disk I/O, you may want to increase the number of workers
+    beyond the number of available CPU cores, to increase utilization. On the other hand,
+    you should be wary about not over-committing resources.
```
amotl, line 212:
```suggestion
    beyond the number of available CPU cores, to increase utilization. On the other hand,
    you should be wary about not over-committing resources too much, as it may slow your system down.
```

docs/by-example/sqlalchemy/dataframe.rst
```diff
@@ -0,0 +1,138 @@
...
+    >>> from pandas._testing import makeTimeDataFrame
+    >>> from crate.client.sqlalchemy.support import insert_bulk
+    ...
+    >>> # Define number of records, and chunk size.
+    >>> INSERT_RECORDS = 42
+    >>> CHUNK_SIZE = 8
+    ...
+    >>> # Create a pandas DataFrame, and connect to CrateDB.
+    >>> df = makeTimeDataFrame(nper=INSERT_RECORDS, freq="S")
+    >>> engine = sa.create_engine(f"crate://{crate_host}")
```
marijaselakovic, line 88: Can we add here the full sqlachemy URI: `crate://user:password@crate_host:port`

setup.py
```diff
@@ -68,6 +68,7 @@ def read(path):
               'zope.testrunner>=5,<7',
               'zc.customdoctests>=1.0.1,<2',
               'createcoverage>=1,<2',
+              'dask',
```
marijaselakovic, line 71: I am not sure but shouldn't we have here `dask[complete]` to install all dask dependencies (e.g., `dask.dataframe`)

marijaselakovic (commented):
The documentation on Dask and Pandas fast insert looks good!

docs/by-example/sqlalchemy/dataframe.rst
marijaselakovic: You might also consider adding Dask Dataframes best practices[^3]

[^3]: https://docs.dask.org/en/stable/dataframe-best-practices.html

docs/by-example/sqlalchemy/dataframe.rst
```diff
@@ -0,0 +1,257 @@
...
+    good value for the ``NPARTITIONS`` argument, based on the capacity and topology of
+    the available compute resources , and based on workload characteristics or policies
+    like peak- vs. balanced- vs. shared-usage. For example, on a machine or cluster fully
+    dedicated to the problem at hand, you may want to use all available processor cores,
+    while on a shared system, this strategy may not be appropriate.
+
+    If you want to dedicate all available compute resources on your machine, you may want
+    to use the number of CPU cores as a value to the ``NPARTITIONS`` argument. You can find
+    out about the available CPU core on your machine, for example by running the ``nproc``
+    command in your terminal.
```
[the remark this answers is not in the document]
amotl: Added as backlog item at https://github.com/crate/crate-python/issues/425#issuecomment-1622612543.

docs/by-example/sqlalchemy/dataframe.rst
amotl: Added as backlog item at https://github.com/crate/crate-python/issues/425#issuecomment-1622612543.

docs/by-example/sqlalchemy/dataframe.rst
amotl: Thanks for your suggestion. It is technically not possible, because this reStructuredText file is not only documentation, but is also part of the test suite, being _tested_ using Python's doctests, in oder to make sure the code examples don't break. Apologies.

setup.py
amotl: This list is only used for testing the functionality, we do not install `pandas` or `dask` together with the library. As such, I think it is fine to use a minimal set of dependencies, which is enough to satisfy the requirements of the test suite.

src/crate/client/sqlalchemy/support.py
amotl: I tried with 4d038ec105fc, but omitted logging the `data` payload.

docs/by-example/sqlalchemy/dataframe.rst
amotl: Good idea. The link has been added.

PR #556, opened 2023-05-26
SQLAlchemy/DDL: Handle `server_default` column schema argument well

### LINKED ISSUE #454: SQLAlchemy/DDL: Support `server_default` column definition option
### Introduction
CrateDB can use functions to generate server-side default values on columns. For example, the `NOW()` function can be used to auto-generate timestamp values.
```sql
crash <<EOL
    DROP TABLE IF EXISTS testdrive;
    CREATE TABLE IF NOT EXISTS "doc"."testdrive" (
      "timestamp" TIMESTAMP DEFAULT NOW() NOT NULL,
      "value" FLOAT
    );
    INSERT INTO testdrive (value) VALUES (42.42);
    INSERT INTO testdrive (value) VALUES (42.42);
    REFRESH TABLE testdrive;
    SELECT * FROM testdrive;
EOL
```

### Objective

It would be sweet if we could make the CrateDB SQLAlchemy dialect support this feature appropriately.

### Proposal

I think the right way is to implement the official SQLAlchemy column definition option `server_default`.
```python
>>> import sqlalchemy as sa
>>> sa.Column("timestamp", sa.DateTime, server_default=sa.text("NOW()"))
```
-- https://docs.sqlalchemy.org/en/20/core/metadata.html#sqlalchemy.schema.Column.params.server_default

### Example
The [collab/sa-column-server-default](https://github.com/crate/crate-python/compare/collab/sa-column-server-default) branch has a corresponding example program.

JanLikar:
### Summary of the changes / Why this is an improvement
CrateDB's SQLAlchemy dialect now handles the `server_default` when generating table DDL.

Fix #454.

### Checklist

### DIFF
--- a/src/crate/client/sqlalchemy/compiler.py
+++ b/src/crate/client/sqlalchemy/compiler.py
@@ -108,7 +108,10 @@ class CrateDDLCompiler(compiler.DDLCompiler):
     def get_column_specification(self, column, **kwargs):
         colspec = self.preparer.format_column(column) + " " + \
             self.dialect.type_compiler.process(column.type)
-        # TODO: once supported add default here
+
+        default = self.get_column_default_string(column)
+        if default is not None:
+            colspec += " DEFAULT " + default
 
         if column.computed is not None:
             colspec += " " + self.process(column.computed)
--- a/src/crate/client/sqlalchemy/tests/create_table_test.py
+++ b/src/crate/client/sqlalchemy/tests/create_table_test.py
@@ -259,3 +259,55 @@ class SqlAlchemyCreateTableTest(TestCase):
 
         with self.assertRaises(sa.exc.CompileError):
             self.Base.metadata.create_all(bind=self.engine)
+
+    def test_column_server_default_text_func(self):
+        class DummyTable(self.Base):
+            __tablename__ = 't'
+            pk = sa.Column(sa.String, primary_key=True)
+            a = sa.Column(sa.DateTime, server_default=sa.text("now()"))
+
+        self.Base.metadata.create_all(bind=self.engine)
+        fake_cursor.execute.assert_called_with(
+            ('\nCREATE TABLE t (\n\t'
+             'pk STRING NOT NULL, \n\t'
+             'a TIMESTAMP DEFAULT now(), \n\t'
+             'PRIMARY KEY (pk)\n)\n\n'), ())
+
+    def test_column_server_default_string(self):
+        class DummyTable(self.Base):
+            __tablename__ = 't'
+            pk = sa.Column(sa.String, primary_key=True)
+            a = sa.Column(sa.String, server_default="Zaphod")
+
+        self.Base.metadata.create_all(bind=self.engine)
+        fake_cursor.execute.assert_called_with(
+            ('\nCREATE TABLE t (\n\t'
+             'pk STRING NOT NULL, \n\t'
+             'a STRING DEFAULT \'Zaphod\', \n\t'
+             'PRIMARY KEY (pk)\n)\n\n'), ())
+
+    def test_column_server_default_func(self):
+        class DummyTable(self.Base):
+            __tablename__ = 't'
+            pk = sa.Column(sa.String, primary_key=True)
+            a = sa.Column(sa.DateTime, server_default=sa.func.now())
+
+        self.Base.metadata.create_all(bind=self.engine)
+        fake_cursor.execute.assert_called_with(
+            ('\nCREATE TABLE t (\n\t'
+             'pk STRING NOT NULL, \n\t'
+             'a TIMESTAMP DEFAULT now(), \n\t'
+             'PRIMARY KEY (pk)\n)\n\n'), ())
+
+    def test_column_server_default_text_constant(self):
+        class DummyTable(self.Base):
+            __tablename__ = 't'
+            pk = sa.Column(sa.String, primary_key=True)
+            answer = sa.Column(sa.Integer, server_default=sa.text("42"))
+
+        self.Base.metadata.create_all(bind=self.engine)
+        fake_cursor.execute.assert_called_with(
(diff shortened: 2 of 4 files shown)


src/crate/client/sqlalchemy/tests/create_table_test.py
```diff
@@ -232,3 +232,55 @@ class DummyTable(self.Base):
             a = sa.Column(Geopoint, crate_index=False)
         with self.assertRaises(sa.exc.CompileError):
             self.Base.metadata.create_all(bind=self.engine)
+
+    def test_column_server_default(self):
```
amotl, line 236: Just a small suggestion on matters of "naming things" .
```suggestion
    def test_column_server_default_text_func(self):
```

src/crate/client/sqlalchemy/tests/create_table_test.py
```diff
@@ -232,3 +232,55 @@ class DummyTable(self.Base):
...
+            a = sa.Column(sa.DateTime, server_default=sa.func.now())
+
+        self.Base.metadata.create_all(bind=self.engine)
+        fake_cursor.execute.assert_called_with(
+            ('\nCREATE TABLE t (\n\t'
+             'pk STRING NOT NULL, \n\t'
+             'a TIMESTAMP DEFAULT now(), \n\t'
+             'PRIMARY KEY (pk)\n)\n\n'), ())
+
+    def test_column_server_default_constant(self):
```
amotl, line 275: Dito.
```suggestion
    def test_column_server_default_text_constant(self):
```

amotl (approved):
Dear Jan,

thanks a stack for submitting this patch. Other than a suggestion about reStructuredText syntax, and maybe a slight naming change on the test methods, I think it will be good to go.

With kind regards,
Andreas.

JanLikar:
> Other than a suggestion about reStructuredText syntax, and maybe a slight naming change on the test methods, I think it will be good to go.
I applied the fixes and squashed the commits. To be honest, I've been using Markdown so much lately I didn't even realize this is rst, thanks for spotting the issue.

src/crate/client/sqlalchemy/tests/create_table_test.py
```diff
@@ -232,3 +232,55 @@ class DummyTable(self.Base):
...
+        class DummyTable(self.Base):
+            __tablename__ = 't'
+            pk = sa.Column(sa.String, primary_key=True)
+            a = sa.Column(sa.DateTime, server_default="Zaphod")
+
+        self.Base.metadata.create_all(bind=self.engine)
+        fake_cursor.execute.assert_called_with(
+            ('\nCREATE TABLE t (\n\t'
+             'pk STRING NOT NULL, \n\t'
+             'a TIMESTAMP DEFAULT \'Zaphod\', \n\t'
```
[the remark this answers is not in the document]
JanLikar: Nice catch. Fixed.

src/crate/client/sqlalchemy/compiler.py
```diff
@@ -107,7 +107,10 @@ class CrateDDLCompiler(compiler.DDLCompiler):
     def get_column_specification(self, column, **kwargs):
         colspec = self.preparer.format_column(column) + " " + \
             self.dialect.type_compiler.process(column.type)
-        # TODO: once supported add default here
+
+        default = self.get_column_default_string(column)
+        if default is not None:
```
sanchezfdezjavier, line 112: Do we intentionally want to intentionally ignore empty values like `[]`, `""`, `{}`.  Why is the reason behind not using?
```suggestion
        if default:
```
Trying to understand, thanks! 😁

seut: @sanchezfdezjavier 
Not sure if I correctly understand your questions, any value beside `None` will evaluate to `True`.
```
'' is not None => True
[] is not None => True
{} is not None => True
False is not None => True
```

While using `if default:` it would evaluate to `False` for all the above examples, see also https://docs.python.org/3/library/stdtypes.html#truth.

amotl: Hi Javi,

`get_column_default_string()` probably returns a string, so in the case of non-empty strings, your suggestion might also work in this case.

But a) @seut is generally right, and b) explicitly testing against `is not None` also feels safer for me, because it does not suffer from any implicit bool-casting flaws.

With kind regards,
Andreas.

src/crate/client/sqlalchemy/tests/create_table_test.py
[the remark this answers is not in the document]
amotl: Hi again. After invoking CI, it looks like the test case `test_column_server_default_string` needs corresponding adjustments to compensate this fix.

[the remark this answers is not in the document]
JanLikar: Hi, sorry about that, I would have run the tests before pushing, but I was traveling without access to a dev environment. 

I'm home now and the tests pass for me, so I think we're good to go.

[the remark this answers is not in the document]
amotl: Dear Jan,

no worries about that hiccup, and welcome back. Now that I've just merged GH-555, it is my turn to apologize for the inconvenience that you may need to rebase your patch now.

If you could also add a changelog item like 7b868010624, the patch will be perfect and complete.

With kind regards,
Andreas.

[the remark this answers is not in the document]
JanLikar: Andreas, thank you for your kind words! I rebased and pushed my changes, luckily the conflicts were relatively mild.

src/crate/client/sqlalchemy/compiler.py
sanchezfdezjavier: Awesome, that makes sense. Thanks for clarifying!

amotl:
Merged. Thanks, and have a good weekend.

Issue #361, opened 2020-07-16
Inserting Datetime aware returns an error. 


MarkRatFelt:
I'm trying to insert some DateTime data to CrateDB, but I'm getting an error. 

The `datetime` I'm trying to insert looks so: 

![Screen Shot 2020-07-16 at 16 25 41](https://user-images.githubusercontent.com/18442189/87683167-0661c200-c781-11ea-8f10-1a4e38399abd.png)

The column is a `Timestamp with time zone` column. 

But I get this error: 

```python
  File "/Users/chuckaguilar/PycharmProjects/postgres-cratedb-migration-tool/venv/lib/python3.7/site-packages/crate/client/http.py", line 82, in default
    delta = o - self.epoch
TypeError: can't subtract offset-naive and offset-aware datetimes
```

But the `CrateJsonEncoder` (`self.epoch`) is naive: 

![Screen Shot 2020-07-16 at 16 28 44](https://user-images.githubusercontent.com/18442189/87683507-73755780-c781-11ea-901b-0b249503557c.png)

`tzinfo is None`

So... it won't work. Either I make my datetime naive, or the `self.epoch` gotta be aware.

MarkRatFelt:
It'd work converting it to a string via `datetime.strftime("%Y-%m-%dT%H:%M:%S.%f%z")`

thanks, @autophagy  :)

PR #557, opened 2023-06-26
Allow handling timezone-aware datetime values when inserting or updating

amotl:
### Problem
As reported at GH-361, when inserting timezone-aware datetime values, the `CrateJsonEncoder` fails with
```python
TypeError: can't subtract offset-naive and offset-aware datetimes
```

This patch intends to fix it.

### Reference
- https://github.com/daq-tools/kotori/pull/148#discussion_r1237793643

### DIFF
--- a/CHANGES.txt
+++ b/CHANGES.txt
@@ -8,6 +8,9 @@ Unreleased
 - SQLAlchemy DDL: Allow turning off column store using ``crate_columnstore=False``.
   Thanks, @fetzerms.
 
+- SQLAlchemy DDL: Allow setting ``server_default`` on columns to enable
+  server-generated defaults. Thanks, @JanLikar.
+
 
 2023/04/18 0.31.1
 =================
@@ -16,7 +19,6 @@ Unreleased
   SQLAlchemy 2.0 by adding the new ``insert_returning`` and ``update_returning`` flags
   in the CrateDB dialect.
 
-- SQLAlchemy DDL: Allow setting ``server_default`` on columns to enable server-generated defaults.
 
 2023/03/30 0.31.0
 =================
--- a/CHANGES.txt
+++ b/CHANGES.txt
@@ -11,6 +11,8 @@ Unreleased
 - SQLAlchemy DDL: Allow setting ``server_default`` on columns to enable
   server-generated defaults. Thanks, @JanLikar.
 
+- Allow handling "aware" datetime values with time zone info when inserting or updating.
+
 
 2023/04/18 0.31.1
 =================
--- a/CHANGES.txt
+++ b/CHANGES.txt
@@ -11,7 +11,7 @@ Unreleased
 - SQLAlchemy DDL: Allow setting ``server_default`` on columns to enable
   server-generated defaults. Thanks, @JanLikar.
 
-- Allow handling "aware" datetime values with time zone info when inserting or updating.
+- Allow handling datetime values tagged with time zone info when inserting or updating.
 
 
 2023/04/18 0.31.1
--- a/docs/by-example/client.rst
+++ b/docs/by-example/client.rst
@@ -142,8 +142,9 @@ Refresh locations:
 Updating Data
 =============
 
-Both when inserting or updating data, values for ``TIMESTAMP`` columns can be obtained
-in different formats. Both literal strings and datetime objects are supported.
+Values for ``TIMESTAMP`` columns can be obtained as a string literal, ``date``,
+or ``datetime`` object. If it contains timezone information, it is converted to
+UTC, and the timezone information is discarded.
 
     >>> import datetime as dt
     >>> timestamp_full = "2023-06-26T09:24:00.123+02:00"
--- a/docs/by-example/client.rst
+++ b/docs/by-example/client.rst
@@ -139,6 +139,24 @@ Refresh locations:
 
     >>> cursor.execute("REFRESH TABLE locations")
 
+Updating Data
+=============
+
+Both when inserting or updating data, values for ``TIMESTAMP`` columns can be obtained
+in different formats. Both literal strings and datetime objects are supported.
+
+    >>> import datetime as dt
+    >>> timestamp_full = "2023-06-26T09:24:00.123+02:00"
+    >>> timestamp_date = "2023-06-26"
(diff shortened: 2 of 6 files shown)


CHANGES.txt
```diff
@@ -8,6 +8,11 @@ Unreleased
 - SQLAlchemy DDL: Allow turning off column store using ``crate_columnstore=False``.
   Thanks, @fetzerms.
 
+- SQLAlchemy DDL: Allow setting ``server_default`` on columns to enable
+  server-generated defaults.
+
+- Allow handling "aware" datetime values when inserting or updating.
```
matriv, line 14:
```suggestion
- Allow handling datetime values with time zone info when inserting or updating.
```

matriv: Suggestion: maybe we could also an example?

matriv (approved):
LGTM, Left a comment for the changelog

src/crate/client/http.py
```diff
@@ -82,13 +83,17 @@ def super_len(o):
 
 class CrateJsonEncoder(json.JSONEncoder):
 
-    epoch = datetime(1970, 1, 1)
+    epoch_aware = datetime(1970, 1, 1, tzinfo=pytz.UTC)
```
mfussenegger, line 86:
```suggestion
    epoch_aware = datetime(1970, 1, 1, tzinfo=timezone.utc)
```

I think it's available in all python versions we support: https://docs.python.org/3/library/datetime.html?highlight=timezone#datetime.timezone.utc

Could drop the pytz dependency.

amotl: Ah sure, I missed that. Thank you. Adjusted the patch.

CHANGES.txt
amotl: I've added a corresponding example to the documentation at `client.rst`. It is not optimal, but the best thing we can do to show the features, and not destroy the storyline of that documentation section too much.

mfussenegger (commented):
I think it would be worth also extending the data type section in the docs and explain the behavior of datetime with tagged timezone.

One could assume that it persists the timezone information as well, but it's instead converting it to a timestamp with 1970-01-01 as epoch. That's an important distinction

docs/by-example/client.rst
```diff
@@ -139,6 +139,24 @@ Refresh locations:
 
     >>> cursor.execute("REFRESH TABLE locations")
 
+Updating Data
+=============
+
+Both when inserting or updating data, values for ``TIMESTAMP`` columns can be obtained
+in different formats. Both scalar strings and datetime objects are supported.
```
matriv, line 146: sorry, not familiar, what is `scalar string`? (is there one of the examples below?

docs/data-types.rst
```diff
@@ -94,9 +94,20 @@ __ https://crate.io/docs/crate/reference/en/latest/general/ddl/data-types.html#c
...
 
+.. NOTE::
+
+    Values of ``TIMESTAMP`` columns will always be stored using a ``LONG`` type,
+    representing the `Unix time`_ (epoch) timestamp, i.e. number of seconds which
+    have passed since 00:00:00 UTC on Thursday, 1 January 1970.
+
+    This means, when inserting or updating records using timezone-aware Python
+    ``datetime`` objects, timezone information will not be preserved. If you
+    need to store it, you will need to use a separate column.
```
amotl, line 108: @mfussenegger: This paragraph intends to reflect your suggestion. Thank you again.

-- https://crate-python--557.org.readthedocs.build/en/557/data-types.html

docs/by-example/client.rst
```diff
@@ -139,6 +139,24 @@ Refresh locations:
...
+
+    >>> import datetime as dt
+    >>> timestamp_full = "2023-06-26T09:24:00.123+02:00"
+    >>> timestamp_date = "2023-06-26"
```
amotl, line 150: @matriv: Those examples, values of `timestamp_full` and `timestamp_date` types, would represent the "scalar strings" variant I was referring to. When referring to scalar vs. container/aggregate types, I usually follow this guidance. I recognize there may be nuances to it.

> Generally, all basic primitive data types are considered scalar.
>
> -- https://stackoverflow.com/a/10212398

docs/by-example/client.rst
matriv: I was just not familiar myself with this term, I'm used to `literals` to describe them, but really as you prefer.

amotl: It's a good suggestion. I rephrased the statement.

> Both literal strings and datetime objects are supported.

docs/by-example/client.rst
```diff
@@ -139,6 +139,24 @@ Refresh locations:
...
+    >>> datetime_aware = dt.datetime.fromisoformat("2023-06-26T09:24:00.123+02:00")
+    >>> datetime_naive = dt.datetime.fromisoformat("2023-06-26T09:24:00.123")
+    >>> datetime_date = dt.date.fromisoformat("2023-06-26")
+    >>> cursor.execute("UPDATE locations SET date=? WHERE name='Cloverleaf'", (timestamp_full, ))
+    >>> cursor.execute("UPDATE locations SET date=? WHERE name='Cloverleaf'", (timestamp_date, ))
+    >>> cursor.execute("UPDATE locations SET date=? WHERE name='Cloverleaf'", (datetime_aware, ))
+    >>> cursor.execute("UPDATE locations SET date=? WHERE name='Cloverleaf'", (datetime_naive, ))
+    >>> cursor.execute("UPDATE locations SET date=? WHERE name='Cloverleaf'", (datetime_date, ))
```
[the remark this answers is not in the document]
amotl: Thanks, added with 677bef31b440.

docs/by-example/client.rst
```diff
@@ -139,6 +139,24 @@ Refresh locations:
 
     >>> cursor.execute("REFRESH TABLE locations")
 
+Updating Data
+=============
+
+Both when inserting or updating data, values for ``TIMESTAMP`` columns can be obtained
+in different formats. Both literal strings and datetime objects are supported.
```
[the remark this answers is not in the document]
amotl: Text has been changed with 677bef31b440, thanks.

> (e.g. do a update returning)

We may elaborate this on another occasion? I will be all ears to adjust the documentation, in order to improve guidance, but I would need a bit of education first.

PR #559, opened 2023-07-04
SQLAlchemy: Fix SQL statement caching for CrateDB's `OBJECT` type

amotl:
## About

The SQLAlchemy implementation of CrateDB's `OBJECT` type offers indexed access to the instance's content in form of a dictionary. Thus, it must not use `cache_ok = True` on its implementation, i.e. to signal this part of the compiled SQL clause must not be cached.

## Problem

Currently, when using those statements subsequently, based on a corresponding ORM schema definition, the second one will yield the wrong results, because it will use the compiled SQL statement of the first one, as it has been cached wrongly. This flaw has been reported by @faymarie, thank you very much.

```python
class Character(Base):
    __tablename__ = 'characters'
    name = sa.Column(sa.String, primary_key=True)
    age = sa.Column(sa.Integer)
    data = sa.Column(Object)
```

```python
sa.select(Character).where(Character.data['x'] == 1)
sa.select(Character).where(Character.data['y'] == 2)
```

## Details

The accompanying test case verifies that two subsequent `SELECT` statements are translated well, and don't trip on incorrect SQL compiled statement caching. Because I have not been able to catch that error by unit tests based on mocking, this test case has now been introduced as a first integration test to the SQLAlchemy dialect subsystem.

### DIFF
--- a/src/crate/client/sqlalchemy/tests/query_caching.py
+++ b/src/crate/client/sqlalchemy/tests/query_caching.py
@@ -1,0 +1,117 @@
+# -*- coding: utf-8; -*-
+#
+# Licensed to CRATE Technology GmbH ("Crate") under one or more contributor
+# license agreements.  See the NOTICE file distributed with this work for
+# additional information regarding copyright ownership.  Crate licenses
+# this file to you under the Apache License, Version 2.0 (the "License");
+# you may not use this file except in compliance with the License.  You may
+# obtain a copy of the License at
+#
+#   http://www.apache.org/licenses/LICENSE-2.0
+#
+# Unless required by applicable law or agreed to in writing, software
+# distributed under the License is distributed on an "AS IS" BASIS, WITHOUT
+# WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.  See the
+# License for the specific language governing permissions and limitations
+# under the License.
+#
+# However, if you have executed another commercial license agreement
+# with Crate these terms will supersede the license and you may use the
+# software solely pursuant to the terms of the relevant commercial agreement.
+
+from __future__ import absolute_import
+from unittest import TestCase, skipIf
+
+import sqlalchemy as sa
+from sqlalchemy.orm import Session
+from sqlalchemy.sql.operators import eq
+
+from crate.client.sqlalchemy import SA_VERSION, SA_1_4
+from crate.testing.settings import crate_host
+
+try:
+    from sqlalchemy.orm import declarative_base
+except ImportError:
+    from sqlalchemy.ext.declarative import declarative_base
+
+from crate.client.sqlalchemy.types import Object, ObjectArray
+
+
+class SqlAlchemyQueryCompilationCaching(TestCase):
+
+    def setUp(self):
+        self.engine = sa.create_engine(f"crate://{crate_host}")
+        self.metadata = sa.MetaData(schema="testdrive")
+        self.session = Session(bind=self.engine)
+        self.Character = self.setup_entity()
+
+    def setup_entity(self):
+        """
+        Define ORM entity.
+        """
+        Base = declarative_base(metadata=self.metadata)
+
+        class Character(Base):
+            __tablename__ = 'characters'
+            name = sa.Column(sa.String, primary_key=True)
+            age = sa.Column(sa.Integer)
+            data = sa.Column(Object)
+            data_list = sa.Column(ObjectArray)
+
+        return Character
+
+    def setup_data(self):
+        """
+        Insert two records into the `characters` table.
+        """
+        self.metadata.drop_all(self.engine)
+        self.metadata.create_all(self.engine)
+
+        Character = self.Character
+        char1 = Character(name='Trillian', data={'x': 1}, data_list=[{'foo': 1, 'bar': 10}])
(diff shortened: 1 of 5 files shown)


src/crate/client/sqlalchemy/types.py
```diff
@@ -132,7 +132,7 @@ def __eq__(self, other):
 
 
 class _Craty(sqltypes.UserDefinedType):
-    cache_ok = True
+    cache_ok = False
```
amotl, line 135: The effective change is this one. The other code is for testing only.

amotl: It may be advisable to look at SQLAlchemy's implementation for PostgreSQL's [JSONB](http://docs.sqlalchemy.org/en/latest/dialects/postgresql.html#sqlalchemy.dialects.postgresql.JSONB) type, and its [`contains()`](http://docs.sqlalchemy.org/en/latest/dialects/postgresql.html#sqlalchemy.dialects.postgresql.JSONB.Comparator.contains) method, or at the corresponding implementation for MSSQL's [JSON](https://docs.sqlalchemy.org/en/latest/dialects/mssql.html#sqlalchemy.dialects.mssql.JSON) type.

-- https://stackoverflow.com/a/39467149

amotl: I played a bit with the `_Craty` implementation for `OBJECT` types within the `crate.client.sqlalchemy.types` module, trying to make this statement work, which I believe would be correct in that sense, and where I would hope the SQLAlchemy machinery would handle it well, without getting the statement caching wrong.
```python
selectable = sa.select(Character).where(Character.data['x'].as_integer() == 1)
```

However, I got different exceptions, so I think it will need further investigations to make this work. In that manner, I suggest to postpone it to a later iteration, and release the bugfix without further ado.

amotl: GH-560 elaborates on what has been suggested above.

src/crate/client/sqlalchemy/tests/query_caching.py
```diff
@@ -0,0 +1,117 @@
...
+        selectable = sa.select(Character).where(Character.data['y'] == 2)
+        result = self.session.execute(selectable).scalar_one().data
+        self.assertEqual({"y": 2}, result)
+
+    @skipIf(SA_VERSION < SA_1_4, "On SA13, the 'ResultProxy' object has no attribute 'scalar_one'")
+    def test_objectarray_multiple_select(self):
+        """
+        The SQLAlchemy implementation of CrateDB's `ARRAY` type in form of the
+        `ObjectArray`, does *not* offer indexed access to the instance's content.
+        Thus, using `cache_ok = True` on that type should be sane, and not mess
```
matriv, line 105: but now we disable cache for both cases, no? and it's not configurable by the user?

amotl: Caching of compiled statements is only turned off for `OBJECT` types, implemented on behalf of `_Craty`. See https://github.com/crate/crate-python/pull/559#pullrequestreview-1512359974.

matriv: Apologies if I'm still confused, but `Character` is still a custom type, I don't get when this cache is used and when not.

amotl: > SQLAlchemy as of version 1.4 includes a [SQL compilation caching facility](https://docs.sqlalchemy.org/en/20/core/connections.html#sql-caching) which will allow Core and ORM SQL constructs to cache their stringified form, along with other structural information used to fetch results from the statement, allowing the relatively expensive string compilation process to be skipped when another structurally equivalent construct is next used. This system relies upon functionality that is implemented for all SQL constructs, including objects such as [Column](https://docs.sqlalchemy.org/en/20/core/metadata.html#sqlalchemy.schema.Column), [select()](https://docs.sqlalchemy.org/en/20/core/selectable.html#sqlalchemy.sql.expression.select), and [TypeEngine](https://docs.sqlalchemy.org/en/20/core/type_api.html#sqlalchemy.types.TypeEngine) objects, to produce a cache key which fully represents their state to the degree that it affects the SQL compilation process.

-- https://docs.sqlalchemy.org/en/20/faq/performance.html ff.

amotl: Currently, accessing the `OBJECT` implementation `_Craty` using dict-like indexing breaks that machinery.

seut (approved):
👍 but please squash as commented, thanks.

src/crate/client/sqlalchemy/tests/query_caching.py
```diff
@@ -0,0 +1,117 @@
...
+        self.setup_data()
+        Character = self.Character
+
+        selectable = sa.select(Character).where(Character.data['x'] == 1)
+        result = self.session.execute(selectable).scalar_one().data
+        self.assertEqual({"x": 1}, result)
+
+        selectable = sa.select(Character).where(Character.data['y'] == 2)
+        result = self.session.execute(selectable).scalar_one().data
+        self.assertEqual({"y": 2}, result)
```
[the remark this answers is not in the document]
amotl: You are right. Commits have been squashed now. Thanks.

matriv (approved):
Thx!

Issue #560, opened 2023-07-04
SQLAlchemy: Investigate whether and how to build upon the `JSON` type implementation


amotl:
## About

GH-559 revealed that the SQLAlchemy implementation for supporting CrateDB's `OBJECT` type has drawbacks when accessing the attributes of the dictionary facade by key, i.e. indexing into the dictionary using the classical `Entity.field["attribute"]` notation. 

## Thoughts

Here, we are trying to outline a possible alternative way of implementing support for that, which may more closely align to PostgreSQL's JSON type support. While CrateDB itself does not provide support for the JSON|JSONB types, the same SQLAlchemy _interface_ could still be leveraged to access CrateDB's `OBJECT` type in the same spirit. At least, this is what I am assuming for now.

## Details

Please inspect SQLAlchemy's [`sqltypes.JSON`](https://github.com/sqlalchemy/sqlalchemy/blob/rel_2_0_17/lib/sqlalchemy/sql/sqltypes.py#L2161-L2725) type implementation, which is provided as a facade for vendor-specific JSON types, which are currently PostgreSQL, MySQL, SQLite, and Microsoft SQL Server. Each of them has their own specializations, so I think the chance to re-use some code from the generic type variants for implementing CrateDB's `OBJECT` type might actually be possible.

### New interface

The new interface to index operations is based on [data casters](https://docs.sqlalchemy.org/en/20/core/type_basics.html#sqlalchemy.types.JSON).

> Index operations, i.e. those invoked by calling upon the expression using the Python bracket operator as in ``some_column['some key']``, return an expression object whose type defaults to [`JSON`](https://docs.sqlalchemy.org/en/20/core/type_basics.html#sqlalchemy.types.JSON) by default, so that further JSON-oriented instructions may be called upon the result type.
>
> However, it is likely more common that an index operation is expected to return a specific scalar element, such as a string or integer. In order to provide access to these elements in a backend-agnostic way, a series of data casters are provided. [...]
>
> See section »Casting JSON Elements to Other Types« at the [SQLAlchemy JSON type](https://docs.sqlalchemy.org/en/20/core/type_basics.html#sqlalchemy.types.JSON) documentation.

### Code examples

When adhering to the new interface, the same statement would be written slightly different.

#### Before
```python
sa.select(Character).where(Character.data['x'] == 1)
```

#### After
```python
sa.select(Character).where(Character.data['x'].as_integer() == 1)
```

amotl:
A bit of progress with GH-561.

[...]

PR #561, opened 2023-07-05
SQLAlchemy: Use JSON type adapter for implementing CrateDB's `OBJECT`

amotl:
## About

Following up on GH-560, by using the generalized parts of SQLAlchemy's infrastructure for implementing JSON types for a few databases, the CrateDB dialect will be able to leverage the same interfaces, like data casters, thus improving compatibility.

#### Example
```python
sa.select(Entity).where(Entity.field['attribute'].as_integer() == 42)
```

## Details

SQLAlchemy's `sqltypes.JSON` provides a facade for vendor-specific JSON types. Since it supports JSON SQL operations, it only works on backends that have an actual JSON type, which are currently PostgreSQL, MySQL, SQLite, and Microsoft SQL Server.

This patch starts leveraging the same infrastructure, thus bringing corresponding interfaces to the CrateDB dialect. The major difference is that it will not actually do any JSON marshalling, but propagate corresponding data structures 1:1, because within CrateDB's SQL, `OBJECT`s do not need to be serialized into JSON strings before transfer.

## Resources

- https://docs.sqlalchemy.org/en/20/core/type_basics.html#sqlalchemy.types.JSON

### DIFF
--- a/CHANGES.txt
+++ b/CHANGES.txt
@@ -15,6 +15,8 @@ Unreleased
 
 - SQLAlchemy: Fix SQL statement caching for CrateDB's ``OBJECT`` type.
 
+- SQLAlchemy: Refactor ``OBJECT`` type to use SQLAlchemy's JSON type infrastructure.
+
 
 2023/04/18 0.31.1
 =================
--- a/src/crate/client/sqlalchemy/types.py
+++ b/src/crate/client/sqlalchemy/types.py
@@ -18,6 +18,7 @@
 # However, if you have executed another commercial license agreement
 # with Crate these terms will supersede the license and you may use the
 # software solely pursuant to the terms of the relevant commercial agreement.
+import warnings
 
 import sqlalchemy.types as sqltypes
 from sqlalchemy.sql import operators, expression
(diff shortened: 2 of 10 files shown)


src/crate/client/sqlalchemy/types.py
```diff
@@ -131,24 +132,31 @@ def __eq__(self, other):
...
+    none_as_null = False
 
-    class Comparator(sqltypes.TypeEngine.Comparator):
 
-        def __getitem__(self, key):
-            return default_comparator._binary_operate(self.expr,
-                                                      operators.getitem,
-                                                      key)
+# Designated name to refer to. `Object` is too ambiguous.
+ObjectType = MutableDict.as_mutable(ObjectTypeImpl)
```
amotl, line 144: SQLAlchemy also offers a `MutableDict` implementation. It looks like using that would resolve some other specializations of the CrateDB dialect, which are a bit anomalous, as we discovered on behalf of the discussion at https://github.com/sqlalchemy/sqlalchemy/discussions/5915.

However, it did not work out of the box, SQLAlchemy's `MutableDict`, or associated code paths, did not seem to support _nested_ object access well. Maybe indeed this is one of the essential specializations the CrateDB dialect has to offer.

This topic will need to be re-visited. /cc GH-425

src/crate/client/sqlalchemy/types.py
```diff
@@ -131,24 +132,31 @@ def __eq__(self, other):
         return dict.__eq__(self, other)
 
 
-class _Craty(sqltypes.UserDefinedType):
+class ObjectTypeImpl(sqltypes.UserDefinedType, sqltypes.JSON):
+
+    __visit_name__ = "OBJECT"
+
     cache_ok = False
```
amotl, line 139: We still need to use `cache_ok = False`. It does not improve the situation wrt. GH-559.

CHANGES.txt
```diff
@@ -15,6 +15,8 @@ Unreleased
 
 - SQLAlchemy: Fix SQL statement caching for CrateDB's ``OBJECT`` type.
 
+- SQLAlchemy: Refactor ``OBJECT`` type to use SQLAlchemy's JSON type machinery.
```
matriv, line 18: I'd personally prefer `infrastructure` or `functionality` over `machinery`.

matriv (approved):
Thank you! Looks good, but please wait for the rest of the reviewers.

CHANGES.txt
amotl: Thanks. Updated the changelog item.

Issue #562, opened 2023-07-05
SQLAlchemy: Investigate if .ilike() should render as ILIKE


hlcianfagna:
`.ilike()` is currently rendering as `lower() LIKE lower()`
Using CrateDB's native ILIKE operator could allow to benefit from a more efficient implementation in the backend in the future.

amotl:
Good idea. The corresponding documentation of SQLAlchemy can be inspected at [^1].

> [`ColumnOperators.ilike()`](https://docs.sqlalchemy.org/en/20/core/sqlelement.html#sqlalchemy.sql.expression.ColumnOperators.ilike):
>
> Case insensitive LIKE makes use of the SQL `lower()` function on a generic backend. On the PostgreSQL backend it will use `ILIKE`.

[^1]: https://docs.sqlalchemy.org/en/20/core/operators.html#string-comparisons

seut:
Implemented by #564.

PR #564, opened 2023-07-06
SQLAlchemy DQL: Use CrateDB's native `ILIKE` operator

amotl:
## About

At GH-562, @hlcianfagna suggested to use CrateDB's native `ILIKE` operator instead of using SA's generic implementation `lower() LIKE lower()`.

## Details

This patch brings in the corresponding visitor methods from the `PGDialect` to the `CrateDialect` implementation.

### DIFF
--- a/src/crate/client/sqlalchemy/compiler.py
+++ b/src/crate/client/sqlalchemy/compiler.py
@@ -244,6 +244,40 @@ class CrateCompiler(compiler.SQLCompiler):
             self.process(element.right, **kw)
         )
 
+    def visit_ilike_case_insensitive_operand(self, element, **kw):
+        """
+        Use native `ILIKE` operator, like PostgreSQL's `PGCompiler`.
+        """
+        return element.element._compiler_dispatch(self, **kw)
+
+    def visit_ilike_op_binary(self, binary, operator, **kw):
+        """
+        Use native `ILIKE` operator, like PostgreSQL's `PGCompiler`.
+
+        Do not implement the `ESCAPE` functionality, because it is not
+        supported by CrateDB.
+        """
+        if binary.modifiers.get("escape", None) is not None:
+            raise NotImplementedError("Unsupported feature: ESCAPE is not supported")
+        return "%s ILIKE %s" % (
+            self.process(binary.left, **kw),
+            self.process(binary.right, **kw),
+        )
+
+    def visit_not_ilike_op_binary(self, binary, operator, **kw):
+        """
+        Use native `ILIKE` operator, like PostgreSQL's `PGCompiler`.
+
+        Do not implement the `ESCAPE` functionality, because it is not
+        supported by CrateDB.
+        """
+        if binary.modifiers.get("escape", None) is not None:
+            raise NotImplementedError("Unsupported feature: ESCAPE is not supported")
+        return "%s NOT ILIKE %s" % (
+            self.process(binary.left, **kw),
+            self.process(binary.right, **kw),
+        )
+
     def limit_clause(self, select, **kw):
         """
         Generate OFFSET / LIMIT clause, PostgreSQL-compatible.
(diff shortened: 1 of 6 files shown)


src/crate/client/sqlalchemy/compiler.py
```diff
@@ -244,6 +244,41 @@ def visit_any(self, element, **kw):
...
+        escape = binary.modifiers.get("escape", None)
+
+        return "%s ILIKE %s" % (
+            self.process(binary.left, **kw),
+            self.process(binary.right, **kw),
+        ) + (
+            " ESCAPE " + self.render_literal_value(escape, sqltypes.STRINGTYPE)
+            if escape
+            else ""
+        )
```
amotl, line 266: Does CrateDB support `ESCAPE` in this context?

hlcianfagna: No, 
```
UnsupportedFeatureException[ESCAPE is not supported.]
```
Maybe we should raise a feature request?

How were % and _ handled previously in the `lower() like lower()` implementation?

hlcianfagna (commented):
See https://github.com/crate/crate-python/pull/564#discussion_r1255381935 about the handling of % and _ in the matching string

src/crate/client/sqlalchemy/compiler.py
amotl: I was hoping that you can explain. ;]

amotl: Shall we just remove the corresponding code, in order to be able to bring in the patch, and create an issue at crate/crate to track concerns of the `ESCAPE` operator/keyword? I guess it can also be used on other statements/clauses?

matriv: Yes please, it seems that postgres supports defining a custom escape char seq:
https://www.postgresql.org/docs/15/functions-matching.html
https://stackoverflow.com/questions/196626/escaping-in-a-ilike-query
which is not the case for CrateDB.

matriv (approved):
Thx!

hlcianfagna (commented):
I confirmed that `ilike` accepts an `escape` parameter and if that is used today it renders as the SQL `ESCAPE` clause which results on an `UnsupportedFeatureException` being returned from CrateDB, if the change goes in as is there would be no error but we would be risking incorrect results, maybe we should raise some exception if the `escape` parameter is used?

matriv:
> I confirmed that `ilike` accepts an `escape` parameter and if that is used today it renders as the SQL `ESCAPE` clause which results on an `UnsupportedFeatureException` being returned from CrateDB, if the change goes in as is there would be no error but we would be risking incorrect results, maybe we should raise some exception if the `escape` parameter is used?

Very true, thx! I agree to throw an exception if `escape` is used.

amotl:
@hlcianfagna: Well spotted, thank you very much. I agree to catch the `escape` argument from `**kwargs`, but not ignore it, and raise a synthetic exception instead, mimicking the one which is currently raised by CrateDB when actually sending the `ESCAPE` keyword.

amotl:
Can we find out since which version CrateDB supports the `ILIKE` operator, and if it would be advisable that we need to fall back to the _generic_ SQLAlchemy implementation, when operating on a lower version of CrateDB?

hlcianfagna:
> Can we find out since which version CrateDB supports the `ILIKE` operator, and if it would be advisable that we need to fall back to the _generic_ SQLAlchemy implementation, when operating on a lower version of CrateDB, like, not breaking support for CrateDB 3.x?

Support for `ILIKE` was added in 4.1.0

matriv (approved):

amotl:
> Support for ILIKE was added in 4.1.0.

Thank you @hlcianfagna. I've just added 596f5e000a, in order to implement corresponding code branching based on the CrateDB version in use.

hlcianfagna (approved):

PR #567, opened 2023-07-27
Added automatic conversion of UUIDs to strings.

SStorm:
## About

When transport-encoding SQL parameters to JSON, Python's UUID type has not been taken into consideration. This patch intends to improve the situation.

### DIFF
--- a/src/crate/client/http.py
+++ b/src/crate/client/http.py
@@ -35,6 +35,8 @@ from base64 import b64encode
 from time import time
 from datetime import datetime, date, timezone
 from decimal import Decimal
+from uuid import UUID
+
 from urllib3 import connection_from_url
 from urllib3.connection import HTTPConnection
 from urllib3.exceptions import (
@@ -86,7 +88,7 @@ class CrateJsonEncoder(json.JSONEncoder):
     epoch_naive = datetime(1970, 1, 1)
 
     def default(self, o):
-        if isinstance(o, Decimal):
+        if isinstance(o, (Decimal, UUID)):
             return str(o)
         if isinstance(o, datetime):
             if o.tzinfo is not None:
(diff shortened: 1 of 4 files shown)


src/crate/client/http.py
```diff
@@ -88,6 +90,8 @@ class CrateJsonEncoder(json.JSONEncoder):
     def default(self, o):
         if isinstance(o, Decimal):
             return str(o)
+        if isinstance(o, UUID):
+            return str(o)
```
amotl, line 94: Might be a slight performance improvement, summing up when sending many requests.

```suggestion
        if isinstance(o, (Decimal, UUID)):
            return str(o)
```

amotl (approved):
Thank you again for addressing my suggestions. LGTM, but I also added @matriv to receive a second perspective.

src/crate/client/test_http.py
```diff
@@ -287,6 +290,18 @@ def test_socket_options_contain_keepalive(self):
         )
         client.close()
 
+    @patch(REQUEST, autospec=True)
+    def test_uuid_serialization(self, request):
+        client = Client(servers="localhost:4200")
+        request.return_value = fake_response(200)
+
+        uid = uuid.uuid4()
+        client.sql('insert into my_table (str_col) values (?)', (uid,))
```
matriv, line 299: out of curiosity: is this trailing `,` of `(uid,)` required? (I see it in other tests as well.

matriv (approved):
LGTM, thx! left just a question out of my curiosity.

src/crate/client/test_http.py
SStorm: That turns it into a tuple in Python, as the client requires an iterable ;) 
```python
>>> type(("test"))
<class 'str'>
>>> type(("test",))
<class 'tuple'>

```

matriv: Great, thx!

PR #582, opened 2023-09-28
SQLAlchemy: Improve DDL compiler to ignore unsupported features

amotl:
## About

Adjust SQLAlchemy's DDL compiler to ignore foreign key and uniqueness constraints, because CrateDB doesn't support them.

## Background

At recent work about unlocking MLflow and LangChain to work together with CrateDB [^1][^2], we needed to add some patches/workarounds.

Both of those to be integrated here are pretty straight-forward and non-invasive, and just mask corresponding SQL features not supported by CrateDB anyway, in order run applications with typical data schemas. Other dialects essentially to the same on their corresponding weak spots.

P.S.: I am intending to bring in this patch before running another release, where users are currently waiting for.

/cc @hlcianfagna, @hammerhead, @surister

[^1]: https://github.com/crate-workbench/mlflow-cratedb
[^2]: https://github.com/crate-workbench/langchain

### DIFF
--- a/CHANGES.txt
+++ b/CHANGES.txt
@@ -13,7 +13,8 @@ Unreleased
   certificate details are immanent, like no longer accepting the long
   deprecated ``commonName`` attribute. Instead, going forward, only the
   ``subjectAltName`` attribute will be used.
-- SQLAlchemy: Improve DDL compiler to ignore foreign key constraints
+- SQLAlchemy: Improve DDL compiler to ignore foreign key and uniqueness
+  constraints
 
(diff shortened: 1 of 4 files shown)


matriv (commented):
Looks good, but how about logging some warning at least?
Additionally I think some NOTE entry in documentation is necessary to inform users as the DDLs would silently succeed and it's hidden from them that those features are unsupported.

amotl:
Good idea. I added 154d77a298051, to emit corresponding warnings, as you suggested.

matriv (approved):
Thx! Since a msg is logged, I think it's enough.

Issue #578, opened 2023-09-14
SQLAlchemy: Expecting `IntegrityError`, but `ProgrammingError` is raised


amotl:
## About

At https://github.com/crate-workbench/mlflow-cratedb/issues/9, we discovered that the CrateDB SQLAlchemy dialect raises a `ProgrammingError` exception when the server raises a `DuplicateKeyException`. However, SQLAlchemy-based applications seem to expect the `IntegrityError` exception type instead, see [sqlalchemy_store.py](https://github.com/mlflow/mlflow/blob/6de2829a28898001aabcc681750d03c5cd291399/mlflow/store/tracking/sqlalchemy_store.py#L731-L739).

## Details

mlflow-cratedb [has a monkeypatch](https://github.com/crate-workbench/mlflow-cratedb/blob/v0.1.1/mlflow_cratedb/patch/crate_python.py#L48-L75), which augments the `_raise_for_status` function to emit an `IntegrityError` correspondingly.

## Proposal

Add this to the code base, untangled from the monkeypatch, together with an accompanying test case, if not present.

amotl:
This has been implemented with GH-583, and we did not receive any complaints so far.

PR #583, opened 2023-09-28
Improve exception handling: Properly raise `IntegrityError` exceptions

amotl:
## About

When receiving `DuplicateKeyException` errors from CrateDB, SQLAlchemy should raise corresponding `IntegrityError` exceptions instead of `ProgrammingError`, because applications expect it this way.

## Details

This patch evaluates the area preliminarily, mainly whether it will work by using CrateDB's own variants of the canonical DBAPI exceptions, and whether SQLAlchemy will honor that properly.

## References

- GH-578


matriv (approved):
Can we have a test case for this?
(Codecov also complaints for line 202 not covered by a test)

matriv (approved):

PR #584, opened 2023-09-28
SQLAlchemy: Ignore SQL's "FOR UPDATE" clause

### LINKED ISSUE #577: SQLAlchemy: Ignore `FOR UPDATE` clauses
## About

At https://github.com/crate-workbench/mlflow-cratedb/issues/7, we discovered that CrateDB does not support `SELECT ... FOR UPDATE` statements.

## Details

mlflow-cratedb [has a monkeypatch](https://github.com/crate-workbench/mlflow-cratedb/blob/v0.1.1/mlflow_cratedb/patch/crate_python.py#L31-L45), which adds a visitor to the CrateDB SQLAlchemy dialect to just omit that clause.

## Proposal

The monkeypatch needs to be converted into a real patch, might be supported by a corresponding software tests to verify the `FOR UPDATE` clause does not get rendered into the SQL statement, and a bit of documentation may also be added - one sentence is probably enough.

## Followup

Determine if this is an applicable candidate to raise on the crate/crate issue tracker.

surister:
Pretty straightforward, now queries like `select(table).with_for_update()` will not generate a `FOR UPDATE` statement.

Resolves #577.
## Checklist

 - [x] Link to issue this PR refers to (if applicable): 

### DIFF
--- a/src/crate/client/sqlalchemy/compiler.py
+++ b/src/crate/client/sqlalchemy/compiler.py
@@ -309,3 +309,10 @@ class CrateCompiler(compiler.SQLCompiler):
         Generate OFFSET / LIMIT clause, PostgreSQL-compatible.
         """
         return PGCompiler.limit_clause(self, select, **kw)
+
+    def for_update_clause(self, select, **kw):
+        # CrateDB does not support the `INSERT ... FOR UPDATE` clause.
+        # See https://github.com/crate/crate-python/issues/577.
+        warnings.warn("CrateDB does not support the 'INSERT ... FOR UPDATE' clause, "
+                      "it will be omitted when generating SQL statements.")
+        return ''
--- a/src/crate/client/sqlalchemy/tests/compiler_test.py
+++ b/src/crate/client/sqlalchemy/tests/compiler_test.py
@@ -43,7 +43,7 @@ from crate.client.test_util import ParametrizedTestCase
 from crate.testing.settings import crate_host
 
 
-class SqlAlchemyCompilerTest(ParametrizedTestCase):
+class SqlAlchemyCompilerTest(ParametrizedTestCase, ExtraAssertions):
 
     def setUp(self):
         self.crate_engine = sa.create_engine('crate://')
@@ -256,6 +256,34 @@ class SqlAlchemyCompilerTest(ParametrizedTestCase):
             mock.call(mock.ANY, 'INSERT INTO mytable (name) VALUES (?), (?)', ('foo_2', 'foo_3'), None),
             mock.call(mock.ANY, 'INSERT INTO mytable (name) VALUES (?)', ('foo_4', ), None),
         ])
+
+    def test_for_update(self):
+        """
+        Verify the `CrateCompiler.for_update_clause` method to
+        omit the clause, since CrateDB does not support it.
+        """
+
+        with warnings.catch_warnings(record=True) as w:
+
+            # By default, warnings from a loop will only be emitted once.
+            # This scenario tests exactly this behaviour, to verify logs
+            # don't get flooded.
+            warnings.simplefilter("once")
+
+            selectable = self.mytable.select().with_for_update()
+            _ = str(selectable.compile(bind=self.crate_engine))
+
+            selectable = self.mytable.select().with_for_update()
+            statement = str(selectable.compile(bind=self.crate_engine))
+
+        # Verify SQL statement.
+        self.assertEqual(statement, "SELECT mytable.name, mytable.data \nFROM mytable")
+
+        # Verify if corresponding warning is emitted, once.
+        self.assertEqual(len(w), 1)
+        self.assertIsSubclass(w[-1].category, UserWarning)
+        self.assertIn("CrateDB does not support the 'INSERT ... FOR UPDATE' clause, "
+                      "it will be omitted when generating SQL statements.", str(w[-1].message))
 
 
 FakeCursor = MagicMock(name='FakeCursor', spec=Cursor)
--- a/CHANGES.txt
+++ b/CHANGES.txt
@@ -17,6 +17,7 @@ Unreleased
   constraints
 - DBAPI: Properly raise ``IntegrityError`` exceptions instead of
   ``ProgrammingError``, when CrateDB raises a ``DuplicateKeyException``.
+- SQLAlchemy: Ignore SQL's ``FOR UPDATE`` clause. Thanks, @surister.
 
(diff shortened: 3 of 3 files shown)


src/crate/client/sqlalchemy/compiler.py
```diff
@@ -292,3 +292,8 @@ def limit_clause(self, select, **kw):
         Generate OFFSET / LIMIT clause, PostgreSQL-compatible.
         """
         return PGCompiler.limit_clause(self, select, **kw)
+
+    def for_update_clause(self, select, **kw):
+        # CrateDB does not support currently 'FOR UPDATE' statements.
+        # See https://github.com/crate/crate-python/issues/577
+        return ''
```
amotl, line 299: I am always not sure if empty string or `None` should be returned here. It could actually be so so. See also GH-582 for comparison. Can you find out more about this detail?

amotl (changes requested):
Thank you very much!

src/crate/client/sqlalchemy/compiler.py
surister: The thing is that `for_update_clause` is later used in https://github.com/sqlalchemy/sqlalchemy/blob/f96da1a8985dfb9303eef41135044c82420d08cc/lib/sqlalchemy/sql/compiler.py#L4919 in string concatenation, returning None would raise a `Type error`

This could be 'patched' if we override `SQLCompiler._compose_select_body` but then it might change this trivial change into changing and overriding some private parts of the class.

amotl: Thanks for the elaboration. I think "empty string" is the right choice here, then. Mind adding a changelog item, so we can merge without further ado?

amotl: Mind adding a changelog item, so we can merge without further ado?

matriv (commented):
Thx a lot for the PR @surister!
I have the same consideration with https://github.com/crate/crate-python/pull/582. `SELECT... FOR UPDATE` has transactional semantics for traditional RDBMS which CRATEDB doesn't support, shouldn't we somehow make inform the user about an unsupported query (log warning message) or at least add this in docs?

src/crate/client/sqlalchemy/compiler.py
amotl, line 299: Good point about the user warning, @matriv. I've just added a corresponding improvement with 154d77a298051.

I think it will be good like this. If you insist on adding it also to the documentation, I can take this, so we do not need to bother @surister about finding a good spot ;].

amotl:
Dear @surister,

after bringing in GH-582 and GH-583, I am going to wrap this up, if you don't have any objections. Users are waiting for another release, and my intention was only to bring in a few small maintenance updates and little improvements like this.

Your contribution was a perfect fit for that, thank you again.

With kind regards,
Andreas.

src/crate/client/sqlalchemy/compiler.py
amotl: I've amended the patch by adding a similar warning like on the other one.

src/crate/client/sqlalchemy/compiler.py
```diff
@@ -309,3 +309,10 @@ def limit_clause(self, select, **kw):
         Generate OFFSET / LIMIT clause, PostgreSQL-compatible.
         """
         return PGCompiler.limit_clause(self, select, **kw)
+
+    def for_update_clause(self, select, **kw):
+        # CrateDB does not support currently 'FOR UPDATE' statements.
```
seut, line 314: `currently` may imply that we will support it in future while its not sure if this will ever be supported
```suggestion
        # CrateDB does not support 'FOR UPDATE' statements.
```

seut (approved):

src/crate/client/sqlalchemy/compiler.py
```diff
@@ -309,3 +309,10 @@ def limit_clause(self, select, **kw):
...
+        # See https://github.com/crate/crate-python/issues/577
+        warnings.warn("CrateDB does not support the 'FOR UPDATE' clause, "
+                      "they will be omitted when generating SQL statements.")
```
amotl, line 317:
```suggestion
                      "it will be omitted when generating SQL statements.")
```

src/crate/client/sqlalchemy/tests/compiler_test.py
```diff
@@ -257,6 +257,34 @@ def test_insert_manyvalues(self):
...
+            statement = str(selectable.compile(bind=self.crate_engine))
+
+        # Verify SQL statement.
+        self.assertEqual(statement, "SELECT mytable.name, mytable.data \nFROM mytable")
+
+        # Verify if corresponding warning is emitted, once.
+        self.assertEqual(len(w), 1)
+        self.assertIsSubclass(w[-1].category, UserWarning)
+        self.assertIn("CrateDB does not support the 'FOR UPDATE' clause, "
+                      "they will be omitted when generating SQL statements.", str(w[-1].message))
```
amotl, line 286:
```suggestion
                      "it will be omitted when generating SQL statements.", str(w[-1].message))
```

matriv (approved):
LGTM. Thank you both @surister and @amotl

src/crate/client/sqlalchemy/tests/compiler_test.py
```diff
@@ -257,6 +257,34 @@ def test_insert_manyvalues(self):
             mock.call(mock.ANY, 'INSERT INTO mytable (name) VALUES (?)', ('foo_4', ), None),
         ])
 
+    def test_for_update(self):
+        """
+        Verify the `CrateCompiler.for_update_clause` method
+        it will be omitted since CrateDB does not support it.
```
amotl, line 263:
```suggestion
        Verify the `CrateCompiler.for_update_clause` method
        to omit the clause, since CrateDB does not support it.
```

src/crate/client/sqlalchemy/compiler.py
amotl: Thanks. Fixed.

src/crate/client/sqlalchemy/compiler.py
amotl: Fixed.

src/crate/client/sqlalchemy/tests/compiler_test.py
amotl: Dito.

src/crate/client/sqlalchemy/tests/compiler_test.py
amotl: Fixed.

Issue #533, opened 2023-03-06
SQLAlchemy/DDL: Support custom `crate_autogenerate_uuid` column definition option


amotl:
### Introduction

CrateDB's `gen_random_text_uuid()` function generates UUID values suitable to be used as auto-generated primary key values, instead of PostgreSQL's `SERIAL` type.
```sql
crash <<EOL
  DROP TABLE IF EXISTS testdrive;
  CREATE TABLE IF NOT EXISTS "doc"."testdrive" (
    "x" TEXT DEFAULT gen_random_text_uuid() NOT NULL,
    "y" INT
  );
  INSERT INTO testdrive (y) VALUES (42);
  INSERT INTO testdrive (y) VALUES (42);
  REFRESH TABLE testdrive;
  SELECT * FROM testdrive;
EOL
```

### Objective

It would be sweet if we could make the CrateDB SQLAlchemy dialect support this feature appropriately.

### Proposal

I think the right way is to implement a custom SQLAlchemy column definition like that:
```python
>>> import sqlalchemy as sa
>>> sa.Column("x", sa.String, primary_key=True, crate_autogenerate_uuid=True)
```

> Support custom column DDL within the generation of CREATE TABLE statements,
> by using the compiler extension documented in [Custom SQL Constructs and 
> Compilation Extension](https://docs.sqlalchemy.org/en/20/core/compiler.html) to extend [`CreateColumn`](https://docs.sqlalchemy.org/en/20/core/ddl.html#sqlalchemy.schema.CreateColumn).
>
> -- https://docs.sqlalchemy.org/en/20/core/ddl.html#sqlalchemy.schema.CreateColumn

### References
- https://github.com/crate/crate-clients-tools/issues/48

amotl:
GH-585 will resolve this by adding a corresponding section to the documentation, to use the `gen_random_text_uuid()` scalar function [^1], which can also be used within an SQL DDL statement, in order to automatically assign random identifiers to newly inserted records on the server side. In this spirit, it is suitable to be used as a ``PRIMARY KEY`` constraint for SQLAlchemy.

[^1]: Available since CrateDB 4.5.0.

PR #585, opened 2023-09-28
Documentation: Add section about using `gen_random_text_uuid` as auto-PK

amotl:
## About

CrateDB 4.5.0 added the `gen_random_text_uuid()` scalar function, which can also be used within an SQL DDL statement, in order to automatically assign random identifiers to newly inserted records on the server side. In this spirit, it is suitable to be used as a ``PRIMARY KEY`` constraint for SQLAlchemy.

This patch adds a corresponding documentation section how to use it with SQLAlchemy.

## References

- Resolves GH-533
- Replaces GH-580

/cc @surister

### DIFF
--- a/docs/sqlalchemy.rst
+++ b/docs/sqlalchemy.rst
@@ -300,6 +300,41 @@ would translate into the following declarative model:
     >>> log.id
     ...
 
+
+Auto-generated primary key
+..........................
+
+CrateDB 4.5.0 added the :ref:`gen_random_text_uuid() <crate-reference:scalar-gen_random_text_uuid>`
+scalar function, which can also be used within an SQL DDL statement, in order to automatically
+assign random identifiers to newly inserted records on the server side.
+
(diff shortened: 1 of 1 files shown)


docs/sqlalchemy.rst
```diff
@@ -300,6 +300,41 @@ would translate into the following declarative model:
...
+assign random identifiers to newly inserted records on the server side.
+
+In this spirit, it is suitable to be used as a ``PRIMARY KEY`` constraint for SQLAlchemy.
+
+A table schema like this
+
+.. code-block:: sql
+
+   CREATE TABLE "doc"."items" (
+     "id" STRING DEFAULT gen_random_text_uuid() NOT NULL,
```
seut, line 318:
```suggestion
     "id" STRING DEFAULT gen_random_text_uuid() NOT NULL PRIMARY KEY,
```

seut (approved):
lgtm otherwise

docs/sqlalchemy.rst
amotl: Fixed, thanks.

PR #591, opened 2023-10-23
Use `verlib2.Version` for comparing versions

amotl:
## About

For not duplicating corresponding code any longer, there is the [verlib2 package](https://pypi.org/project/verlib2/) now, which includes the same canonical code from `packaging.version` as before, see GH-513.


seut (dismissed):
👍

seut (commented):
great! but it seems not to work as expected as a lot of test fail?

src/crate/client/_pep440.py
```diff
@@ -1,501 +0,0 @@
...
-    @property
-    def is_postrelease(self):
-        return bool(self._version.post)
-
-    @property
-    def version(self) -> tuple:
-        """
-        PATCH: Return version tuple for backward-compatibility.
-        """
-        return self._version.release
```
amotl, line 403: @seut: This little polyfill needs to be added to verlib2, to make it compatible.

amotl: The snippet to gain backwards-compatibility has been added with verlib2 0.2.0. It resolves the issue of failing test cases.

seut (approved):

PR #593, opened 2023-11-06
CI: Use CrateDB 5.4.5

amotl:
## About

On CI, use a more recent CrateDB version for the regular runs.

## NB
Download location is https://cdn.crate.io/downloads/releases/crate-5.4.4.tar.gz. Correct or outdated/legacy? /cc @seut

### DIFF
--- a/.github/workflows/tests.yml
+++ b/.github/workflows/tests.yml
@@ -19,12 +19,12 @@ jobs:
     strategy:
       matrix:
         os: ['ubuntu-latest', 'macos-latest']
-        cratedb-version: ['5.2.2']
         python-version: ['3.7', '3.8', '3.9', '3.10', '3.11', '3.12']
+        cratedb-version: ['5.4.5']
         sqla-version: ['<1.4', '<1.5', '<2.1']
         pip-allow-prerelease: ['false']
 
-        # To save resources, only use the most recent Python version on macOS.
+        # To save resources, only use the most recent Python versions on macOS.
         exclude:
           - os: 'macos-latest'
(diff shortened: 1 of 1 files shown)


seut:
> Download location is https://cdn.crate.io/downloads/releases/crate-5.4.4.tar.gz. Correct or outdated/legacy? /cc @seut

Yes this location is the legacy download url used when CrateDB **did not** bundle the JDK.
Would be great if we could switch to using os/arch related download url's in future.

.github/workflows/tests.yml
```diff
@@ -19,12 +19,12 @@ jobs:
     strategy:
       matrix:
         os: ['ubuntu-latest', 'macos-latest']
-        cratedb-version: ['5.2.2']
         python-version: ['3.7', '3.8', '3.9', '3.10', '3.11', '3.12']
+        cratedb-version: ['5.4.4']
```
seut, line 23: As `5.4.5` was declared stable today, can we use it instead?
```suggestion
        cratedb-version: ['5.4.5']
```

.github/workflows/tests.yml
```diff
@@ -39,7 +39,7 @@ jobs:
         include:
           - os: 'ubuntu-latest'
             python-version: '3.12'
-            cratedb-version: '5.2.2'
+            cratedb-version: '5.4.4'
```
seut, line 42:
```suggestion
            cratedb-version: '5.4.5'
```

seut (approved):
👍 I'd suggest to use `5.4.5` instead.

PR #595, opened 2023-11-09
async for crate

rventer-hartree:
## Checklist

 - [ ] Link to issue this PR refers to (if applicable): 
 - [ ] [CLA](https://crate.io/community/contribute/cla/) is signed

### DIFF
--- a/aiocrate/__init__.py
+++ b/aiocrate/__init__.py
@@ -1,0 +1,44 @@
+# Copyright 2022 Amethyst Reese
+# Licensed under the MIT license
+
+"""asyncio bridge to the standard sqlite3 module"""
+
+from sqlite3 import (  # pylint: disable=redefined-builtin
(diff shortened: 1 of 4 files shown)


amotl:
Dear @rventer-hartree,

there is another patch in the making which will bring async support for the CrateDB SQLAlchemy dialect.

- https://github.com/crate-workbench/sqlalchemy-cratedb/pull/11

With kind regards,
Andreas.

amotl:
Hi again,

because we do not see any `async def` definitions within this patch, and also don't see how any other asynchronous transport library is being used, while urllib3 does not provide any support for it, we are wondering which kind of difference that patch makes for you?

Could you elaborate a bit more what your goals have been? Maybe we can find a common ground to proceed on this.

With kind regards,
Andreas.

PR #597, opened 2023-11-22
CI: Fix running nightly jobs with CrateDB nightly

amotl:
## Problem

Apparently, GH-594 had a flaw when selecting CrateDB nightly, as we can see [on the nightly jobs](https://github.com/crate/crate-python/actions/workflows/nightly.yml), which probably started failing since this change.

## Solution

This patch adjusts the `sed` command correspondingly, so the nightly job will pick up the CrateDB nightly tarball for Linux again.


mkleen (approved):

matriv:
This is because the nightly one is only on the root folder and not under os specific dirs.

PR #603, opened 2024-01-12
Permit `urllib3.Timeout` instances for defining timeout values

amotl:
## About

Permit `urllib3.Timeout` instances for defining timeout values. Using that, you can configure both the socket `connect` and `read` timeout settings, in seconds.

## Problem
When defining the `timeout` value as `urllib3.Timeout` instance,
```python
HttpClient(crate_host, timeout=urllib3.Timeout(connect=0.01, read=None))
```
this error is raised:
```python
float() argument must be a string or a real number, not 'Timeout'
```

## References
- Needed by: https://github.com/crate/crash/pull/421

### DIFF
--- a/CHANGES.txt
+++ b/CHANGES.txt
@@ -5,6 +5,10 @@ Changes for crate
 Unreleased
 ==========
 
+- Permit ``urllib3.Timeout`` instances for defining timeout values.
+  This way, both ``connect`` and ``read`` socket timeout settings can be
+  configured. The unit is seconds.
+
 
 2023/09/29 0.34.0
 =================
--- a/docs/by-example/http.rst
+++ b/docs/by-example/http.rst
@@ -199,10 +199,21 @@ timeout exception:
     {...}
     >>> http_client.close()
 
-It's possible to define a HTTP timeout in seconds on client instantiation, so
-an exception is raised when the timeout is reached:
+It is possible to define a HTTP timeout in seconds when creating a client
+object, so an exception is raised when the timeout expires:
 
     >>> http_client = HttpClient(crate_host, timeout=0.01)
+    >>> http_client.sql('select fib(32)')
+    Traceback (most recent call last):
+    ...
+    crate.client.exceptions.ConnectionError: No more Servers available, exception from last server: ...
+    >>> http_client.close()
+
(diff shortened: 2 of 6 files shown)


docs/by-example/http.rst
```diff
@@ -199,8 +199,8 @@ timeout exception:
     {...}
     >>> http_client.close()
 
-It's possible to define a HTTP timeout in seconds on client instantiation, so
-an exception is raised when the timeout is reached:
+It is possible to define a HTTP timeout in seconds when creating a client
+object, so an exception is raised when the timeout expired:
```
matriv, line 203:
```suggestion
object, so an exception is raised when the timeout expires:
```

CHANGES.txt
```diff
@@ -5,6 +5,9 @@ Changes for crate
 Unreleased
 ==========
 
+- Stopped decoding the ``timeout`` parameter strictly as scalar to also
```
matriv, line 8: Please use a user facing message here, so what has changed for the user, not the internal implementation details.

surister (approved):

CHANGES.txt
amotl: Thanks. I've amended both the changelog item, and the commit message according to your suggestions.

CHANGES.txt
```diff
@@ -5,6 +5,10 @@ Changes for crate
 Unreleased
 ==========
 
+- Permit ``urllib3.Timeout`` instances for defining timeout values.
+  Using that, you can configure both the socket ``connect`` and ``read``
```
matriv, line 9: Maybe reads better:
```
This way, both ``connect`` and ``read`` socket timeout settings can be configured, in seconds.
```

matriv (approved):

CHANGES.txt
amotl: Thanks for your suggestion. I used it 1:1, yet separated the mention of the unit into a separate sentence.

> Permit ``urllib3.Timeout`` instances for defining timeout values.
> This way, both ``connect`` and ``read`` socket timeout settings can be
> configured. The unit is seconds.

mfussenegger (approved):

PR #616, opened 2024-06-10
Remove SQLAlchemy dialect

amotl:
## About
This patch removes the SQLAlchemy dialect from the DB API driver. It lives in the [`sqlalchemy-cratedb`](https://github.com/crate/sqlalchemy-cratedb) repository now.

## Details
This branch will be used for testing purposes right away, and must only be merged when sqlalchemy-cratedb is ready, i.e. when it is considered to be good enough to not cause too much friction for downstream users.

## References
- https://github.com/crate/roadmap/issues/85

/cc @karynzv, @hlcianfagna, @hammerhead, @tomach, @proddata, @surister, @seut, @simonprickett


amotl:
> [...] must only be merged when sqlalchemy-cratedb is ready, i.e. when it is considered to be good enough to not cause too much friction for downstream users.

[sqlalchemy-cratedb](https://github.com/crate/sqlalchemy-cratedb) has been slotted into a few downstream projects already, and CI runs on them are succeeding across the board.

[PyPI Downloads](https://pepy.tech/project/sqlalchemy-cratedb/)

The amount of traffic on that dependency is 2k/month in June 2024 already, so we are considering it stable enough for a switchover, paired with corresponding instructions how to [migrate from `crate.client`](https://cratedb.com/docs/sqlalchemy-cratedb/migrate-from-crate-client.html).

PR #618, opened 2024-06-13
Chore: Stop building universal wheels, being in Python 3 lands only

amotl:
## Problem

The wheel file name of `crate` still indicates the package would be compatible with Python 2 [^1], which isn't the case. Other packages are doing it right [^2].

[^1]: https://pypi.org/project/crate/0.35.2/#files
[^2]: https://pypi.org/project/cratedb-toolkit/#files

### DIFF
--- a/setup.cfg
+++ b/setup.cfg
@@ -1,5 +1,2 @@
-[wheel]
-universal = 1
-
 [flake8]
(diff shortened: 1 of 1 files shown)


seut (approved):
Good catch

surister (approved):

PR #624, opened 2024-06-17
CI: Use CrateDB `nightly` also for regular testing

amotl:
## About
CrateDB version was fixed to 5.4.5. This patch intends to use ~~`latest`~~ `nightly`.

## Thoughts
`latest` will probably not work, because the test layer used here only understands specific version numbers, and `nightly`?

### DIFF
--- a/.github/workflows/tests.yml
+++ b/.github/workflows/tests.yml
@@ -19,7 +19,7 @@ jobs:
       matrix:
         os: ['ubuntu-latest', 'macos-latest']
         python-version: ['3.7', '3.8', '3.9', '3.10', '3.11', '3.12']
-        cratedb-version: ['5.7.2']
+        cratedb-version: ['nightly']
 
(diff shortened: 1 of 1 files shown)


matriv:
Do you have any concern to test on CI with nightly? It could potentially help us catch issues of incompatibilities between CrateDB and python, before a release.

amotl:
Thanks. Let's go for nightly right away, also for regular testing. CrateDB Nightly is used for crate-python nightly CI jobs anyway, so let's synchronize this.

amotl:
404 Not Found @ https://cdn.crate.io/downloads/releases/cratedb/x64_mac/crate-latest.tar.gz -- what's that?

-- https://github.com/crate/crate-python/actions/runs/9553310174/job/26331846066?pr=624#step:4:230

matriv (approved):

amotl:
This is an alternative that works now.

- GH-662

PR #635, opened 2024-06-26
Configure DB API interface attribute `threadsafety = 1`: Threads may share the module, but not connections

amotl:
## About
Relating to the [section about thread safety in PEP 0249](https://peps.python.org/pep-0249/#threadsafety) ...

- The driver currently configures `threadsafety = 2`, which means "Threads may share the module and connections."
- Now, it will configure `threadsafety = 1`, which means "Threads may share the module, but not connections."

## Discussion

@hlcianfagna asked:

> When using the python driver, is it ok to share a connection with multiple threads each opening their own cursor or should each thread have its own connection?

@amotl said:

> Thread-per-connection sounds reasonable.

@mfussenegger said:

> I think it currently relies on Python GIL, and afaik if you'd share a connection with the `multiprocessing` module across processes, things would break. So, it could be worth changing that to 1.

@hlcianfagna said:

> I would say the same, should we document it?

## References
- [Python Driver: Using multiple cursors parallel](https://community.cratedb.com/t/python-crate-using-multiple-cursors-parallel/1788)

/cc @hlcianfagna, @hammerhead, @surister, @proddata, @simonprickett

### DIFF
--- a/CHANGES.txt
+++ b/CHANGES.txt
@@ -8,6 +8,8 @@ Unreleased
 - The SQLAlchemy dialect has been split off into the `sqlalchemy-cratedb`_
   package. See `Migrate from crate.client to sqlalchemy-cratedb`_ to learn
   about necessary migration steps.
+- Configured DB API interface attribute ``threadsafety = 1``, which signals
+  "Threads may share the module, but not connections."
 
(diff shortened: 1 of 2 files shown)


amotl:
@hlcianfagna said:

> I would say the same, should we document it?

Let me know if you think we should also add a dedicated section within the documentation. I think it could be worth it, if we don't refer to relevant thread safety details yet. I didn't check yet, though.

seut (approved):

surister (approved):

PR #648, opened 2024-10-02
Add `error_trace` to string representation of an Error

seut:
If the `error_trace` payload is available, add it to the string representation of the Error class.

This way, any layer using the driver, e.g. our SqlAlchemy Dialect, can display the error trace without to handle concrete exceptions. Additionally, handling the concrete exception and reading the `self.error_trace` variable won't be always possible (e.g. in generic SqlAlchemy abstractions).
If the error_trace is available is already depending on the concrete `connect(error_trace=True)` attribute and thus in control of the user.

### DIFF
--- a/CHANGES.txt
+++ b/CHANGES.txt
@@ -10,6 +10,8 @@ Unreleased
   about necessary migration steps.
 - Configured DB API interface attribute ``threadsafety = 1``, which signals
   "Threads may share the module, but not connections."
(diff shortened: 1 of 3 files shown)


amotl (approved):
Thank you.

amotl:
That other patch validates the improvement on the downstream SQLAlchemy dialect.

- https://github.com/crate/sqlalchemy-cratedb/pull/161

PR #649, opened 2024-10-02
Add `BulkResponse` wrapper for improved decoding of HTTP bulk responses

amotl:
## About
CrateDB HTTP bulk responses include `rowcount=` items, either signalling if a bulk operation succeeded or failed.

- success means `rowcount=1`
- failure means `rowcount=-2`

https://cratedb.com/docs/crate/reference/en/latest/interfaces/http.html#error-handling

## References
The code is coming from CrateDB Toolkit, but is generally usable beyond there.

- https://github.com/crate/cratedb-toolkit/blob/v0.0.27/cratedb_toolkit/io/core.py#L17-L77
- https://github.com/crate/cratedb-toolkit/pull/270


src/crate/client/test_result.py
```diff
@@ -0,0 +1,44 @@
...
+        self.assertEqual(result, [{"rowcount": -2}, {"rowcount": -2}])
+
+        # Verify decoded response.
+        bulk_response = BulkResponse(invalid_records, result)
+        self.assertEqual(bulk_response.failed_records, invalid_records)
+
+        cursor.execute("REFRESH TABLE foobar;")
+        cursor.execute("SELECT * FROM foobar;")
+        result = cursor.fetchall()
+        self.assertEqual(sorted(result), [[1], [2], [3]])
```
amotl, line 41: Currently, CrateDB seems to fail the second batch `[(4, ), ("Hotzenplotz", )]` completely. Is that normal?

src/crate/client/tests.py
amotl, line 1: `tests.py` is the entrypoint for zope.testing to discover test cases. Many utility functions had to be refactored away from here, in order to avoid circular imports.

src/crate/client/result.py
```diff
@@ -0,0 +1,66 @@
...
+    @cached_property
+    def failed_records(self) -> t.List[t.Dict[str, t.Any]]:
+        """
+        Compute list of failed records.
+
+        CrateDB signals failed inserts using `rowcount=-2`.
+
+        https://cratedb.com/docs/crate/reference/en/latest/interfaces/http.html#error-handling
+        """
+        if self.records is None or self.results is None:
```
matriv, line 37: I'd rather raise errors (even separate errors in this case. (meaning if self.records is "None" one error, if self.results is "None", a separate error.

src/crate/client/test_result.py
matriv: If there is syntax error, or error in analysis (e.g.: mismatching values with data types) the whole bulk fails before starting executing any of the inserts. See the issue and my comment: https://github.com/crate/crate/issues/16683#issuecomment-2376499787 for details.

src/crate/client/test_result.py
```diff
@@ -0,0 +1,48 @@
...
+
+        # Import at runtime is on purpose, to permit skipping.
+        from crate.client.result import BulkResponse
+
+        connection = client.connect(crate_host)
+        cursor = connection.cursor()
+
+        # Run SQL DDL and a first valid INSERT operation.
+        cursor.execute("CREATE TABLE foobar (id INTEGER);")
+        _ = cursor.executemany("INSERT INTO foobar (id) VALUES (?)", [(1, ), (2, ), (3, )])
```
matriv, line 28: To really check bulk execution errors, you need for example to add a PK on id, and then try to insert duplicate ids.

matriv (approved):
Thx for improving the bulk response handling. Left some comments.

src/crate/client/test_result.py
amotl: Thanks!

src/crate/client/test_result.py
amotl: Thank you very much. I will change the test case correspondingly, to better exercise partially succeeding batches.

src/crate/client/result.py
amotl: The callsite at [^1] likes to have the situation handled gracefully, but maybe that is just because I wanted to have it "convenient". Please poke me another time if you think the component should really raise an exception in this situation, and I will be happy to adjust.

[^1]: https://github.com/crate/cratedb-toolkit/blob/v0.0.27/cratedb_toolkit/io/core.py#L147-L149

src/crate/client/test_result.py
amotl: The test case has been changed like you suggested, and it works perfectly well.

src/crate/client/test_result.py
```diff
@@ -0,0 +1,70 @@
...
+        # Run SQL DDL.
+        cursor.execute("CREATE TABLE foobar (id INTEGER PRIMARY KEY, name STRING);")
+
+        # Run a batch insert that is empty.
+        with self.assertRaises(ProgrammingError) as cm:
+            cursor.executemany("INSERT INTO foobar (id, name) VALUES (?, ?)", [])
+        self.assertEqual(
+            cm.exception.message,
+            "SQLParseException[The query contains a parameter placeholder $1, "
+            "but there are only 0 parameter values]")
```
amotl, line 67: I see. Running an empty batch into `executemany()` will already raise an exception here. In this spirit, it is probably sane to also let `BulkResponse` raise an exception?

src/crate/client/result.py
matriv: imho, and if I understand it correctly those 2 shouldn't be none unless something has gone really wrong. So why not handle it properly, catching the exceptions in the toolkit if needed?

src/crate/client/test_result.py
matriv: I'd say so, because it's the case that the batch is failing as a whole before the execution is started.

src/crate/client/test_result.py
matriv: nice!

src/crate/client/result.py
amotl: Improved with 4da893a73720, thanks!

src/crate/client/test_result.py
```diff
@@ -0,0 +1,88 @@
...
+
+        # Run a batch insert that only partially succeeds.
+        invalid_records = [(1, "Hotzenplotz 1"), (1, "Hotzenplotz 2")]
+        result = cursor.executemany("INSERT INTO foobar (id, name) VALUES (?, ?)", invalid_records)
+
+        # Verify CrateDB response.
+        self.assertEqual(result, [{"rowcount": 1}, {"rowcount": -2}])
+
+        # Verify decoded response.
+        bulk_response = BulkResponse(invalid_records, result)
```
matriv, line 38: Looking at this more carefully, I don't like that BulkResponse is something you need con construct manually. Couldn't we directly return it from the insert execution, instead of a list of `BulkResultItem`?

amotl: Hi. I don't think we can do anything like this here, because the Python database driver must adhere to the [Python Database API Specification](https://peps.python.org/pep-0249/), so this is just meant as an optional extension.

amotl: af32409f7925fe, just added, provides a bit of documentation for that extension in the context of bulk operations.

mfussenegger (commented):
Could you elaborate a bit more on the motivation for this?

Seems to me as if getting the info without this is not really that much more difficult.

amotl:
Hi @mfussenegger,

> The code is coming from CrateDB Toolkit, but is generally usable beyond there.
>
> - https://github.com/crate/cratedb-toolkit/blob/v0.0.27/cratedb_toolkit/io/core.py#L17-L77
> - https://github.com/crate/cratedb-toolkit/pull/270

The idea is to have a concise code representation [over here](https://github.com/crate/cratedb-toolkit/blob/v0.0.27/cratedb_toolkit/io/core.py#L144-L151), for the `BulkProcessor` component, by abstracting away the handling of magic numbers from user-side code.

Because CrateDB Toolkit is rather heavy, and I wanted to make it reusable, I wanted to add it elsewhere. Another option would be to use the now separate cratedb-sqlalchemy package. `BulkProcessor` depends on it anyway, but `BulkResponse` doesn't, and would provide a generic little convenience wrapper around CrateDB's special responses to bulk requests.

mfussenegger:
> The idea is to have a concise code representation [over here](https://github.com/crate/cratedb-toolkit/blob/v0.0.27/cratedb_toolkit/io/core.py#L144-L151), for the `BulkProcessor` component, by abstracting away the handling of magic numbers from user-side code.

That kinda confirms my point in that there's not much more code without the BulkResponse wrapper:


```python
 cursor = self.connection.execute(statement=statement, parameters=operation.parameters)
 self.connection.commit()
 cratedb_bulk_result = getattr(cursor.context, "last_executemany_result", None)
 failed_records = 0
 success_count = 0
 for result in cratedb_bulk_result:
     if result["rowcount"] == -2:
         failed_records += 1
     else:
         success_count += 1
 self._metrics.count_success_total += success_count
 self.progress_bar and self.progress_bar.update(n=success_count)
```

This to me would even be more readable as it avoids the additional indirection.

Also, given that the DB API spec says:

> Return values are not defined.

I think we could extend our return value - as long as we can keep it compatible with the existing one.

PR #651, opened 2024-10-02
Chore: Use CrateDB 5.8.3 for testing

amotl:
## About
Maintenance: Just update CrateDB to use the most recent version.

## Details
Unfortunately, the test suite currently can't use neither of "latest", nor "testing".

## References
- GH-624


matriv (approved):
thx, keep in mind we're gonna release 5.8.4 soon.

PR #653, opened 2024-10-07
CI: Verify support on Python 3.13

amotl:
## About
[Python 3.13.0](https://www.python.org/downloads/release/python-3130/) has been released on Oct. 7, 2024. This PR intends to add CI verification. Also, 627fb09ec5e includes a removal of dependency packages not needed here any longer.


.github/workflows/tests.yml
```diff
@@ -17,11 +17,11 @@ jobs:
     runs-on: ${{ matrix.os }}
     strategy:
       matrix:
-        os: ['ubuntu-latest', 'macos-latest']
-        python-version: ['3.7', '3.8', '3.9', '3.10', '3.11', '3.12']
+        os: ['ubuntu-22.04', 'macos-latest']
+        python-version: ['3.7', '3.8', '3.9', '3.10', '3.11', '3.12', '3.13']
```
kneth, line 21: Can 3.7 be removed? I mean, it is not an active version anymore.

kneth (approved):
My comment isn't blocking a merge. Asked mostly out of curiosity.

.github/workflows/tests.yml
surister: I'd say so

surister: or maybe we just keep it just in case, since most people are still on 3.8-3.9, so I assume another bulk will still be at 3.7

amotl: The package is still compatible even with Python 3.6 or even lower versions of Python. Even while we officially don't support it any longer, and despite it's EOL, I don't think we need to drop support on purpose. As many lovely stories go, it might be a lifesaver to some people out there.

surister: Good by me

kneth: As long as it doesn't prevent us from evolving our code base, I have no issues with supported non-active versions.

PR #659, opened 2024-10-31
Testing: Refactor software tests into dedicated directory `tests`

amotl:
## About
Refactor the layout of the software tests: Tests are no longer inline with their modules (which also made them being included into the release packages), but now live in a separate, dedicated directory within the repository, a conventional top-level `tests` folder.

Contrary to the **test cases** that have been refactored, the **testing utilities** within `crate.testing` still exist, and will continue to be shipped.

## Rationale
Before releasing 1.0.0, we would like to get the shape of the repository in order. This patch makes a start on that.

## Preview
https://github.com/crate/crate-python/tree/better-testing/tests

## Details
No other functional changes inside.

## References
- GH-660


kneth (approved):
That's a lot of moving stuff around 😄

amotl:
> That's a lot of moving stuff around 😄

I did it _just for you_ to have a perfect view into what's actually inside, by digesting software tests, now better to explore and read than beforehand.

-- https://github.com/crate/crate-python/tree/master/tests

Note that the repository also includes doctests, that can also help to get an idea about the surface of the library.

-- https://github.com/crate/crate-python/tree/master/docs/by-example
-- https://cratedb.com/docs/python/en/latest/by-example/

PR #660, opened 2024-10-31
QA and CI: Format code using ruff. Validate using ruff and mypy.

amotl:
## About
- Run the ruff code formatter and linter on the code base, and the mypy type checker.
- Invoke the "lint" job also on CI.

## Rationale
Before releasing 1.0.0, we would like to get the shape of the repository in order. This patch is one in a series to support that.

## Details
No other functional changes inside.

## References
- GH-659

### DIFF
--- a/src/crate/client/cursor.py
+++ b/src/crate/client/cursor.py
@@ -18,21 +18,20 @@
 # However, if you have executed another commercial license agreement
 # with Crate these terms will supersede the license and you may use the
 # software solely pursuant to the terms of the relevant commercial agreement.
+import typing as t
+import warnings
 from datetime import datetime, timedelta, timezone
 
-from .converter import DataType
-import warnings
-import typing as t
-
-from .converter import Converter
+from .converter import Converter, DataType
 from .exceptions import ProgrammingError
 
 
-class Cursor(object):
+class Cursor:
     """
     not thread-safe by intention
     should not be shared between different threads
     """
+
     lastrowid = None  # currently not supported
 
     def __init__(self, connection, converter: Converter, **kwargs):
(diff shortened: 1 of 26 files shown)


matriv (approved):

src/crate/client/cursor.py
```diff
@@ -306,12 +300,17 @@ def _to_datetime_with_tz(value: t.Optional[float]) -> t.Optional[datetime]:
     @staticmethod
     def _timezone_from_utc_offset(tz) -> timezone:
         """
-        Convert UTC offset in string format (e.g. `+0530`) into `datetime.timezone` object.
+        UTC offset in string format (e.g. `+0530`) to `datetime.timezone`.
         """
-        assert len(tz) == 5, f"Time zone '{tz}' is given in invalid UTC offset format"
+        # TODO: Remove use of `assert`. Better use exceptions?
```
kneth, line 305: Only if you will catch the exception where you call `_timezone_from_utc_offset`

kneth (approved):
Reformatting always gives you huge PRs.

src/crate/client/cursor.py
amotl: On this spot, I am referring to the fact that the ruff linter discourages the use of `assert` statements in code:
```
S101 Use of `assert` detected
```

That's why it needs to be suppressed now, using `# noqa: S101`.

My proposal would be to rewrite the `assert` statements to constructs like that:
```python
if len(tz) != 5:
    raise ValueError(f"Time zone '{tz}' is given in invalid UTC offset format")
```

I think that would be the correct approach to modernize the code base. wdyt?

kneth: Now it makes sense, and I agree that raising an exception is a better option.

PR #662, opened 2024-10-31
CI: Use CrateDB `nightly` for PRs on Linux, lock version only on macOS

amotl:
## Problem
There are no nightly builds for macOS, so the test matrix has an anomaly. C'est la vie.

## Improvement
By restructuring this spot on the CI configuration, at least we don't need to bump versions manually any longer to prevent falling behind CrateDB development. While this just applies to the Linux slot, it is much better than nothing, and before.


matriv (approved):
thx

Issue #665, opened 2024-11-01
SetuptoolsDeprecationWarning: The namespace_packages parameter is deprecated.


amotl:
## About
When building the package, there is a warning. [PEP 420](https://peps.python.org/pep-0420/) is not exactly a new thing, so this package could also make the transition?

## Details
```
python -m build
```
```
* Building wheel...
/private/var/folders/06/w9pzygdj7vx53n_0l9q_lhph0000gn/T/build-env-xjqhhfwn/lib/python3.13/site-packages/setuptools/dist.py:702: SetuptoolsDeprecationWarning: The namespace_packages parameter is deprecated.
!!

        ********************************************************************************
        Please replace its usage with implicit namespaces (PEP 420).

        See https://setuptools.pypa.io/en/latest/references/keywords.html#keyword-namespace-packages for details.
        ********************************************************************************

!!
  ep.load()(self, ep.name, value)
```
-- https://setuptools.pypa.io/en/latest/references/keywords.html#keyword-namespace-packages

PR #666, opened 2024-11-01
Python: Migrate to use "implicit namespace packages" (PEP 420)

amotl:
## About
... instead of "declared namespaces" for the `crate` namespace package, see [PEP 420](https://peps.python.org/pep-0420/), and [setuptools docs about `namespace_packages`](https://setuptools.pypa.io/en/latest/references/keywords.html#keyword-namespace-packages).

> Historically, there were two methods to create namespace packages. One is the `pkg_resources` style supported by `setuptools` and the other one being `pkgutils` style offered by `pkgutils` module in Python. Both are now considered _deprecated_.
>
> -- [Legacy Namespace Packages](https://setuptools.pypa.io/en/latest/userguide/package_discovery.html#legacy-namespace-packages)

## References
- GH-665

/cc @surister


mfussenegger (approved):

kneth (commented):
LGTM but definitely not a Python expert

PR #667, opened 2024-11-01
Timestamp values: Remove the use of "naive" Python `datetime` objects

amotl:
## About
Python: Remove invocations to deprecated `datetime.utcfromtimestamp()`.
World: Make datetime handling like it's 2024.

## Summary
This is a possible BREAKING CHANGE about returned Python ``datetime`` objects:
> Removed the use of "naive" Python ``datetime`` objects, i.e. instances without ``tzinfo`` attribute set.

## Details
When no ``time_zone`` information is specified when creating a database connection or cursor, ``datetime`` objects will now use Coordinated Universal Time (UTC), like CrateDB is storing timestamp values in this format.

This update is coming from a deprecation of Python's ``datetime.utcfromtimestamp()``, which is effectively also phasing out the use of "naive" timestamp objects in Python, in favor of using timezone-aware objects, also to represent datetimes in UTC. It may be a breaking change for some users of the library that don't expect to receive "aware" ``datetime`` objects from now on.

```python
DeprecationWarning: datetime.datetime.utcfromtimestamp() is deprecated and scheduled for removal in a future version. Use timezone-aware objects to represent datetimes in UTC: datetime.datetime.fromtimestamp(timestamp, datetime.UTC).
```

### DIFF
--- a/src/crate/client/converter.py
+++ b/src/crate/client/converter.py
@@ -24,9 +24,9 @@ Machinery for converting CrateDB database types to native Python data types.
 https://crate.io/docs/crate/reference/en/latest/interfaces/http.html#column-types
 """
 
+import datetime as dt
 import ipaddress
 from copy import deepcopy
-from datetime import datetime
 from enum import Enum
 from typing import Any, Callable, Dict, List, Optional, Union
 
@@ -45,13 +45,13 @@ def _to_ipaddress(
     return ipaddress.ip_address(value)
 
 
-def _to_datetime(value: Optional[float]) -> Optional[datetime]:
+def _to_datetime(value: Optional[float]) -> Optional[dt.datetime]:
     """
     https://docs.python.org/3/library/datetime.html
     """
     if value is None:
         return None
-    return datetime.utcfromtimestamp(value / 1e3)
+    return dt.datetime.fromtimestamp(value / 1e3, tz=dt.timezone.utc)
 
 
 def _to_default(value: Optional[Any]) -> Optional[Any]:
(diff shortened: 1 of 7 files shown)


src/crate/client/converter.py
```diff
@@ -45,13 +45,13 @@ def _to_ipaddress(
...
 
-def _to_datetime(value: Optional[float]) -> Optional[datetime]:
+def _to_datetime(value: Optional[float]) -> Optional[dt.datetime]:
     """
     https://docs.python.org/3/library/datetime.html
     """
     if value is None:
         return None
-    return datetime.utcfromtimestamp(value / 1e3)
+    return dt.datetime.fromtimestamp(value / 1e3, tz=dt.timezone.utc)
```
kneth, line 54: Would it be faster to `value * 1e-3`?

tests/client/test_cursor.py
```diff
@@ -205,7 +205,16 @@ def test_execute_with_converter(self):
...
-                    datetime.datetime(2022, 7, 18, 18, 10, 36, 758000),
+                    datetime.datetime(
+                        2022,
+                        7,
+                        18,
+                        18,
+                        10,
+                        36,
+                        758000,
+                        tzinfo=datetime.timezone.utc,
```
kneth, line 216: Do we need to test other time zones too?

kneth (approved):

src/crate/client/converter.py
amotl: idk.

tests/client/test_cursor.py
amotl: The test suite bears a few other test cases that do something with non-UTC time zones. Do you like them?

https://github.com/crate/crate-python/blob/f0ef825c42865a29084f11de1107814eda7a5acd/tests/client/test_cursor.py#L46-L148

kneth: Thanks! All I need 😄

PR #676, opened 2024-11-13
Chore: Suppress unqualified CodeQL admonitions

amotl:
## About
GitHub's CodeQL flags [1] a few spots unqualified with "Unused global variable" [2].

image

## Solution?
Based on a suggestion [3], this patch attempts to use the `advanced-security/dismiss-alerts` [4] GitHub Action recipe to provide measures to suppress CodeQL flagging by using inline code annotations.

[1] https://github.com/crate/crate-python/security/code-scanning
[2] https://codeql.github.com/codeql-query-help/python/py-unused-global-variable/
[3] https://github.com/github/codeql/issues/11427#issuecomment-1721059096
[4] https://github.com/advanced-security/dismiss-alerts

## References
- https://github.com/advanced-security/dismiss-alerts/issues/118

### DIFF
--- a/.github/workflows/codeql.yml
+++ b/.github/workflows/codeql.yml
@@ -53,6 +53,8 @@ jobs:
           languages: ${{ matrix.language }}
           config-file: ./.github/codeql.yml
           queries: +security-and-quality
+          # run an 'alert-suppression' query
+          packs: "codeql/${{ matrix.language }}-queries:AlertSuppression.ql"
 
       #- name: Autobuild
       #  uses: github/codeql-action/autobuild@v2
@@ -62,4 +64,21 @@ jobs:
           uv pip install --system '.[test]'
 
       - name: Perform CodeQL Analysis
+        id: analyze
         uses: github/codeql-action/analyze@v3
+        with:
+          category: "/language:${{matrix.language}}"
+          # define the output folder for SARIF files
+          output: sarif-results
+
+      # Unlock inline mechanism to suppress CodeQL warnings.
+      # https://github.com/github/codeql/issues/11427#issuecomment-1721059096
+      - name: Dismiss alerts
+        # if: github.ref == 'refs/heads/main'
+        uses: advanced-security/dismiss-alerts@v1
+        with:
+          # specify a 'sarif-id' and 'sarif-file'
+          sarif-id: ${{ steps.analyze.outputs.sarif-id }}
+          sarif-file: sarif-results/${{ matrix.language }}.sarif
+        env:
+          GITHUB_TOKEN: ${{ github.token }}
--- a/src/crate/client/__init__.py
+++ b/src/crate/client/__init__.py
@@ -31,6 +31,7 @@ __all__ = [
 # regex!
 __version__ = "1.0.0"
 
+# codeql[py/unused-global-variable]
 apilevel = "2.0"
 threadsafety = 1
 paramstyle = "qmark"


.github/workflows/codeql.yml
```diff
@@ -62,4 +62,21 @@ jobs:
...
+      # https://github.com/github/codeql/issues/11427#issuecomment-1721059096
+      - name: Dismiss alerts
+        # if: github.ref == 'refs/heads/main'
+        uses: advanced-security/dismiss-alerts@v1
+        with:
+          # specify a 'sarif-id' and 'sarif-file'
+          sarif-id: ${{ steps.analyze.outputs.sarif-id }}
+          sarif-file: sarif-results/${{ matrix.language }}.sarif
+        env:
+          GITHUB_TOKEN: ${{ github.token }}
```
amotl, line 82: Hi @aibaars. Thank you very much for submitting a fix to add the missing `id: analyze`.

amotl: However, it looks like the admonitions that have been opened are not automatically being dismissed again. On the CodeQL workflow run, there is a warning:

### 1 configuration not found
Warning: Code scanning may not have found all the alerts introduced by this pull request, because 1 configuration present on refs/heads/main was not found:

### Actions workflow (codeql.yml)
    ⌛  .github/workflows/codeql.yml:analyze/language:python

-- https://github.com/crate/crate-python/pull/676/checks?check_run_id=32972101671

aibaars: You also need to run an `alert-suppression` query. That is a QL analysis that searches for the `#codeql` comments. According to the example in the readme this can be done as follows:
```
    - name: Initialize CodeQL
      uses: github/codeql-action/init@v3
      with:
        languages: ${{ matrix.language }}
        # run an 'alert-suppression' query
        packs: "codeql/${{ matrix.language }}-queries:AlertSuppression.ql"
```

aibaars: Another problem, you'll likely end up with a bunch of pairs of new and fixed alerts because you are adding the `#codeql` comments on the same line as the alert. CodeScanning uses a hash-based "fingerprint" to track an alert across commits. This hash is calculated from the 100 non-whitespace character counted from the start of the line on which the alert occurs. When you put a suppression comment on the same line, then CodeScanning considers the alert without the comment to be a different one. As a result the old one is closed as "fixed" and a new one is created in the "dismissed" state. 

You can put the comment on the line before the alert to avoid this problem in most cases. This is why the documentation recommends the following style:

```python
# codeql[py/unused-global-variable]
apilevel = "2.0" 
```

However, if you have alerts very close to each other (fewer than 100 characters apart) you still see this behaviour.

amotl: Hi @aibaars,

thank you so much for your elaborate and valuable comments.

I've adjusted the patch to use your suggestions, which resolved the admonitions by @github-advanced-security at https://github.com/crate/crate-python/pull/676#pullrequestreview-2434698664. We guess that will also resolve the items at [security/code-scanning](https://github.com/crate/crate-python/security/code-scanning) after merging.

Excellent!

With kind regards,
Andreas.

amotl: I guess this warning will resolve itself also after merging?

> ### 1 configuration not found
>
> Warning: Code scanning cannot determine the alerts introduced by this pull request, because 1 configuration present on refs/heads/main was not found:
>
> -- https://github.com/crate/crate-python/pull/676/checks?check_run_id=33421035019

Issue #689, opened 2025-01-15
json.dumps takes significant CPU time in data heavy workloads like ETL or streaming


widmogrod:
Hi Team,

I use the crate-python driver during CDC process that writes data to CrateDB.
During benchmarking and optimizations of the process I recoded CPU flamegraphs that show that a lot of CPU time was spent during json serialization https://github.com/crate/crate-python/blob/c6892d5f57104177da1502941cceb73e815fc5ff/src/crate/client/http.py#L337

There are [publicly available benchmarks reporting that json.dumps is very slow](https://dollardhingra.com/blog/python-json-benchmarking/), and changing the library to ujson or  orjson can make a huge difference. I can confirm that swapping json to ujson reduces time spent on serialization by around 30-40%. 

I didn't check how this impacts correctness since the current implementation uses JSON.dumps(cls=) to provide custom transformation logic.

Cheers,
Gabriel

amotl:
Thanks for your suggestion. Switching to orjson on that spot you've outlined seems to work well, as it doesn't break the test suite.

- GH-691

amotl:
If you are using SQLAlchemy, please also check this patch, which will unlock a communication path not using JSON marshalling at all.

- https://github.com/crate/sqlalchemy-cratedb/pull/11

amotl:
@seut: Do you agree to include GH-691 into a release version 2.0.0, as proposed?

@widmogrod: Because we received approval of GH-691, may I kindly ask you to test `crate==2.0.0.dev0` already in a downstream application (actually, as many as possible would be excellent), in order to validate that nothing goes south?

amotl:
@widmogrod: If you want to use it together with SQLAlchemy, you will currently need to install from Git.
```
pip install 'sqlalchemy-cratedb @ git+https://github.com/crate/sqlalchemy-cratedb.git@crate-orjson'
```
Let me know if you also need a dev release of `sqlalchemy-cratedb` on PyPI.

widmogrod:
I used poetry to fetch  `crate = "^2.0.0.dev1"` and run benchmark
This is how a section of the flame chart looks like 
- left - crate = "^1.0.0"
- right - `crate = "^2.0.0.dev1"`

This part of the code is much more faster:
<img width="1336" alt="Image" src="https://github.com/user-attachments/assets/0bdbe7c7-35aa-4c1e-bafa-b1c3fc4b56fb" />

amotl:
### Update
[...]

Issue #652, opened 2024-10-03
Data Type: `NUMERIC`


amotl:
## About
CrateDB gained a new data type [NUMERIC](https://cratedb.com/docs/crate/reference/en/master/general/ddl/data-types.html#numeric-precision-scale). Most probably, we want to support it here in one way or another.

## References
- https://github.com/crate/sqlalchemy-cratedb/issues/163
- https://github.com/crate-workbench/meltano-target-cratedb/issues/52

amotl:
It is probably still the right choice to serialize Decimal values as strings, because there is no other way to marshal them through JSON?

https://github.com/crate/crate-python/blob/8a3bf388e07e463a9ea3f451a4bdb6c64f2242fd/src/crate/client/http.py#L95-L96

/cc @matriv, @seut

matriv:
If we have the same behavior that we encountered here: https://github.com/crate/crate/pull/16843
I agree to go with strings.

amotl:
Thank you. Then, there will be nothing to update or improve, other than the documentation perhaps, also on the Python driver at a relevant spot?

-- https://github.com/crate/crate/pull/16843

As you advise that HTTP/JSON transport may loose precision, may I ask if it would be expectably different when using the PostgreSQL interface?

seut:
> As you advise that HTTP/JSON transport may loose precision, may I ask if it would be expectably different when using the PostgreSQL interface?

Yes, the issue is completely related to JSON only.

matriv:
and to be more precise, http streams the full number, but the `application/json` MIME type directive is the one that truncates digits.

seut:
As CrateDB supports `NUMERIC`/`DECIMAL` types for quite some time (only the support to *store* numeric values was added recently), I wonder what this issue is about, what is missing?

bgunebakan:
Closing this as superseded by #826 new ticket explains the problem precisely with reproduction, root cause, and a concrete proposed fix.

PR #691, opened 2025-01-15
Use `orjson` to improve JSON marshalling performance

amotl:
## About
Using [orjson](https://pypi.org/project/orjson/) for JSON marshalling can yield performance wins. It has been used in CrateDB Toolkit already, in that case its sister library [orjsonl](https://pypi.org/project/orjsonl/). Thanks, @widmogrod.

## References
- GH-689
- https://github.com/crate/cratedb-toolkit/pull/270

## Downstream
- https://github.com/crate/crash/pull/459
- https://github.com/crate/cratedb-toolkit/pull/349
- https://github.com/crate/cratedb-examples/pull/811
- https://github.com/crate/sqlalchemy-cratedb/pull/195
- https://github.com/crate/langchain-cratedb/pull/23
- https://github.com/crate/academy-fundamentals-course/pull/49
- https://github.com/crate/cloud-api/pull/3038
- https://github.com/crate/stevedore/pull/325
- https://github.com/crate/grand-central/pull/189

### DIFF
--- a/src/crate/client/http.py
+++ b/src/crate/client/http.py
@@ -20,6 +20,8 @@
 # software solely pursuant to the terms of the relevant commercial agreement.
 
 
+import calendar
+import datetime as dt
 import heapq
 import io
 import logging
@@ -84,19 +86,35 @@ def super_len(o):
     return None
 
 
-def cratedb_json_encoder(obj: t.Any) -> str:
+epoch_aware = dt.datetime(1970, 1, 1, tzinfo=dt.timezone.utc)
+epoch_naive = dt.datetime(1970, 1, 1)
+
+
+def json_encoder(obj: t.Any) -> t.Union[int, str]:
     """
     Encoder function for orjson, with additional type support.
 
-    - Python's `Decimal` type.
-    - freezegun's `FakeDatetime` type.
+    - Python's `Decimal` type will be serialized to `str`.
+    - Python's `dt.datetime` and `dt.date` types will be
+      serialized to `int` after converting to milliseconds
+      since epoch.
 
     https://github.com/ijl/orjson#default
+    https://cratedb.com/docs/crate/reference/en/latest/general/ddl/data-types.html#type-timestamp
     """
     if isinstance(obj, Decimal):
         return str(obj)
-    elif hasattr(obj, "isoformat"):
-        return obj.isoformat()
+    if isinstance(obj, dt.datetime):
+        if obj.tzinfo is not None:
+            delta = obj - epoch_aware
+        else:
+            delta = obj - epoch_naive
+        return int(
+            delta.microseconds / 1000.0
+            + (delta.seconds + delta.days * 24 * 3600) * 1000.0
+        )
+    if isinstance(obj, dt.date):
+        return calendar.timegm(obj.timetuple()) * 1000
     raise TypeError
 
 
@@ -108,8 +126,12 @@ def json_dumps(obj: t.Any) -> bytes:
     """
     return orjson.dumps(
         obj,
-        default=cratedb_json_encoder,
-        option=(orjson.OPT_NON_STR_KEYS | orjson.OPT_SERIALIZE_NUMPY),
+        default=json_encoder,
+        option=(
+            orjson.OPT_PASSTHROUGH_DATETIME
+            | orjson.OPT_NON_STR_KEYS
+            | orjson.OPT_SERIALIZE_NUMPY
+        ),
     )
 
 
--- a/src/crate/client/http.py
+++ b/src/crate/client/http.py
@@ -20,23 +20,21 @@
 # software solely pursuant to the terms of the relevant commercial agreement.
 
 
-import calendar
 import heapq
 import io
-import json
 import logging
 import os
 import re
 import socket
 import ssl
 import threading
+import typing as t
 from base64 import b64encode
-from datetime import date, datetime, timezone
 from decimal import Decimal
 from time import time
 from urllib.parse import urlparse
-from uuid import UUID
 
+import orjson
 import urllib3
 from urllib3 import connection_from_url
 from urllib3.connection import HTTPConnection
@@ -86,25 +84,33 @@ def super_len(o):
     return None
 
 
-class CrateJsonEncoder(json.JSONEncoder):
-    epoch_aware = datetime(1970, 1, 1, tzinfo=timezone.utc)
-    epoch_naive = datetime(1970, 1, 1)
+def cratedb_json_encoder(obj: t.Any) -> str:
+    """
+    Encoder function for orjson, with additional type support.
 
-    def default(self, o):
-        if isinstance(o, (Decimal, UUID)):
-            return str(o)
-        if isinstance(o, datetime):
-            if o.tzinfo is not None:
-                delta = o - self.epoch_aware
-            else:
-                delta = o - self.epoch_naive
-            return int(
-                delta.microseconds / 1000.0
-                + (delta.seconds + delta.days * 24 * 3600) * 1000.0
-            )
-        if isinstance(o, date):
-            return calendar.timegm(o.timetuple()) * 1000
-        return json.JSONEncoder.default(self, o)
+    - Python's `Decimal` type.
+    - freezegun's `FakeDatetime` type.
+
+    https://github.com/ijl/orjson#default
+    """
+    if isinstance(obj, Decimal):
+        return str(obj)
+    elif hasattr(obj, "isoformat"):
+        return obj.isoformat()
+    raise TypeError
+
+
+def json_dumps(obj: t.Any) -> bytes:
+    """
+    Serialize to JSON format, using `orjson`, with additional type support.
+
+    https://github.com/ijl/orjson
+    """
+    return orjson.dumps(
+        obj,
+        default=cratedb_json_encoder,
+        option=(orjson.OPT_NON_STR_KEYS | orjson.OPT_SERIALIZE_NUMPY),
+    )
 
 
 class Server:
@@ -180,7 +186,7 @@ class Server:
 
 def _json_from_response(response):
     try:
-        return json.loads(response.data.decode("utf-8"))
+        return orjson.loads(response.data)
     except ValueError as ex:
         raise ProgrammingError(
             "Invalid server response of content-type '{}':\n{}".format(
@@ -223,7 +229,7 @@ def _raise_for_status_real(response):
     if response.status == 503:
         raise ConnectionError(message)
     if response.headers.get("content-type", "").startswith("application/json"):
-        data = json.loads(response.data.decode("utf-8"))
+        data = orjson.loads(response.data)
         error = data.get("error", {})
         error_trace = data.get("error_trace", None)
         if "results" in data:
@@ -323,7 +329,7 @@ def _update_pool_kwargs_for_ssl_minimum_version(server, kwargs):
             kwargs["ssl_minimum_version"] = ssl.TLSVersion.MINIMUM_SUPPORTED
 
 
-def _create_sql_payload(stmt, args, bulk_args):
+def _create_sql_payload(stmt, args, bulk_args) -> bytes:
     if not isinstance(stmt, str):
         raise ValueError("stmt is not a string")
     if args and bulk_args:
@@ -334,7 +340,7 @@ def _create_sql_payload(stmt, args, bulk_args):
         data["args"] = args
     if bulk_args:
         data["bulk_args"] = bulk_args
-    return json.dumps(data, cls=CrateJsonEncoder)
+    return json_dumps(data)
 
 
 def _get_socket_opts(
--- a/tests/client/test_http.py
+++ b/tests/client/test_http.py
@@ -318,7 +318,7 @@ class HttpClientTest(TestCase):
         # convert string to dict
         # because the order of the keys isn't deterministic
         data = json.loads(request.call_args[1]["data"])
-        self.assertEqual(data["args"], ["2015-02-28T07:31:40"])
+        self.assertEqual(data["args"], [1425108700000])
         client.close()
 
     @patch(REQUEST, autospec=True)
@@ -329,7 +329,7 @@ class HttpClientTest(TestCase):
         day = dt.date(2016, 4, 21)
         client.sql("insert into users (dt) values (?)", (day,))
         data = json.loads(request.call_args[1]["data"])
-        self.assertEqual(data["args"], ["2016-04-21"])
+        self.assertEqual(data["args"], [1461196800000])
         client.close()
 
     def test_socket_options_contain_keepalive(self):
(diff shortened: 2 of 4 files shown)


tests/client/test_http.py
```diff
@@ -724,10 +724,10 @@ def test_username(self):
 class TestCrateJsonEncoder(TestCase):
     def test_naive_datetime(self):
         data = dt.datetime.fromisoformat("2023-06-26T09:24:00.123")
-        result = json.dumps(data, cls=CrateJsonEncoder)
-        self.assertEqual(result, "1687771440123")
+        result = cratedb_json_encoder(data)
+        self.assertEqual(result, 1687771440123)
```
amotl, line 728: That's apparently the only kind of anomaly that can be observed after also switching JSON loading per `orjson.loads`.

```diff
-"1687771440123"
+1687771440123
```

amotl: It can probably be fixed easily by adjusting the `cratedb_json_encoder` slightly?

seut (approved):

tests/client/test_http.py
amotl: Excellent. I was dearly hoping for such an answer, that it is actually correct behaviour now.

src/crate/client/http.py
```diff
@@ -86,25 +86,31 @@ def super_len(o):
...
+            delta = obj - epoch_aware
+        else:
+            delta = obj - epoch_naive
+        return int(
+            delta.microseconds / 1000.0
+            + (delta.seconds + delta.days * 24 * 3600) * 1000.0
+        )
+    if isinstance(obj, date):
+        return calendar.timegm(obj.timetuple()) * 1000
+    return obj
```
amotl, line 113: In CrateDB Toolkit, I am deriving from the now-pruned traditional CrateJsonEncoder per [CrateJsonEncoderWithNumPy](https://github.com/crate/cratedb-toolkit/blob/a69a9ca3a5819f6d6fb09f941c040a6c772bed4e/cratedb_toolkit/util/sqlalchemy.py#L31-L57).

I am wondering if and how the [relevant extension that accounts for NumPy types](https://github.com/crate/cratedb-toolkit/blob/a69a9ca3a5819f6d6fb09f941c040a6c772bed4e/cratedb_toolkit/util/sqlalchemy.py#L48-L56) should be added to the core implementation instead, in order to make it more versatile, but without adding NumPy as a required dependency.

amotl: > [...] wondering if and how the relevant extension that accounts for NumPy types should be added.

It looks like this will become a non-issue, because [orjson](https://pypi.org/project/orjson/) can handle NumPy types already:

> It serializes [dataclass](https://github.com/ijl/orjson?tab=readme-ov-file#dataclass), [datetime](https://github.com/ijl/orjson?tab=readme-ov-file#datetime), [numpy](https://github.com/ijl/orjson?tab=readme-ov-file#numpy), and [UUID](https://github.com/ijl/orjson?tab=readme-ov-file#uuid) instances natively.

amotl: Do you think we should add relevant software tests to crate-python, which marshal a few examples of NumPy types, while the NumPy package will only become part of the **test dependencies** in crate-python?

I think any other framework naturally would do downstream integration tests on their own behalves, like sqlalchemy-cratedb is doing it, and many others, out of band of the test suite of crate-python.

However, because the relevant fragments of code in focus here are extending the core type marshalling function `cratedb_json_encoder`, it may be applicable to integrate a small fragment of NumPy testing code into the test suite itself, now that we know NumPy types will probably not need any special treatments, just to make sure about it.

amotl (commented):
Hi again. Thanks for the quick review. I've just added a few thoughts about the consequences of removing the traditional `CrateJsonEncoder`.

src/crate/client/http.py
```diff
@@ -86,25 +86,31 @@ def super_len(o):
...
+epoch_naive = datetime(1970, 1, 1)
+
+
+def cratedb_json_encoder(obj):
+    """
+    Encoder function for orjson.
+
+    https://github.com/ijl/orjson#default
+    https://github.com/ijl/orjson#opt_passthrough_datetime
+    """
```
amotl, line 99: Like the [opt_passthrough_dataclass](https://github.com/ijl/orjson?tab=readme-ov-file#opt_passthrough_dataclass) option conveys it:

> Passthrough `dataclasses.dataclass` instances to `default`. This allows customizing their output but is much slower.

... I think the same holds true also for the [opt_passthrough_datetime](https://github.com/ijl/orjson?tab=readme-ov-file#opt_passthrough_datetime) option? JFYI.

src/crate/client/http.py
seut: > Do you think we should add relevant software tests to crate-python, which marshal a few examples of NumPy types, while the NumPy package will only become part of the test dependencies in crate-python?

As you wrote `because [orjson](https://pypi.org/project/orjson/) can handle NumPy types already:` it seems that this feature is purely inside `orjson` so it should be tested there already. I don't see any reason to test the functionality of a dependency inside our code. Do I miss something? ;)

amotl: All right. I just wanted to check (better safe than sorry), I am mostly bringing it up about not missing support for NumPy types.

Now, after [reading up on the docs](https://github.com/ijl/orjson?tab=readme-ov-file#numpy), I can see:

> orjson is faster than all compared libraries at serializing numpy instances. Serializing numpy data requires specifying `option=orjson.OPT_SERIALIZE_NUMPY`.

So, I guess we'd like to enable this option, if you don't have any objections.

src/crate/client/http.py
seut: According to the [docs](https://github.com/ijl/orjson?tab=readme-ov-file#datetime), `orjson` serializes `DateTime` objects to a String format CrateDB understands. I wonder if we need this at all?

amotl: I've added this admonition to the OP.

seut: Same for the `data` type, the string format `"1900-01-02"` is understood by CrateDB (see https://github.com/crate/crate/blob/master/server/src/test/java/io/crate/types/DateTypeTest.java#L45).

Also [UUID](https://github.com/ijl/orjson?tab=readme-ov-file#uuid) is supported and correctly serialized to a valid string. 

So seems like we only need any custom encoder for the `DECIMAL` type?

src/crate/client/http.py
amotl: Enabled per 20e735fcf.

src/crate/client/http.py
amotl: > So seems like we only need any custom encoder for the `DECIMAL` type?

idk. I had created an issue about it over here the other day, if I am getting it right that `DECIMAL` ==  `NUMERIC`?

- https://github.com/crate/crate-python/issues/652

amotl: > According to the [docs](https://github.com/ijl/orjson?tab=readme-ov-file#datetime), `orjson` serializes `DateTime` objects to a String format CrateDB understands. I wonder if we need this at all?

I actually was wondering about it as well. However, I started bringing this in by keeping it as a fact that we'd need a custom decoder. Thanks for bringing it up: Let's try to swap it out and check what CI will tell us.

seut: > if I am getting it right that DECIMAL == NUMERIC

Yes.
I've just [added a comment](https://github.com/crate/crate-python/issues/652#issuecomment-2595082040) to the mentioned issue as it's not really clear to me what is missing to support `decimal` values, it should already work.

amotl: With two more adjustments to test cases, the test suite succeeds also without `OPT_PASSTHROUGH_DATETIME`. If you will confirm this is a good idea, the positive consequence is to trim the new `cratedb_json_encoder` function significantly?

amotl: When adding 03559a9028, the test suite [starts failing](https://github.com/crate/crate-python/actions/runs/12806754738/job/35705886534#step:5:177).

Expectations are that unixtime values are exchanged, while it suddenly became a native Python datetime object.                                                                                                                                                                                                                                                                                                               
```
 AssertionError: datetime.datetime(2023, 6, 26, 9, 24, 0, [62 chars]00))) != 1687764240123
```

src/crate/client/http.py
amotl: https://github.com/crate/cratedb-toolkit/pull/349 shows that NumPy support works well, so we have been able to remove relevant monkeypatches needed before. 💯

src/crate/client/http.py
amotl: Because crate-2.0.0.dev1 works [so excellently](https://github.com/crate/crate-python/pull/691#discussion_r1918115729), I'd tend to skip this within the current iteration on this patch, and possibly add it as an improvement later.

src/crate/client/http.py
```diff
@@ -334,7 +340,11 @@ def _create_sql_payload(stmt, args, bulk_args):
         data["args"] = args
     if bulk_args:
         data["bulk_args"] = bulk_args
-    return json.dumps(data, cls=CrateJsonEncoder)
+    return orjson.dumps(
+        data,
+        default=cratedb_json_encoder,
+        option=(orjson.OPT_PASSTHROUGH_DATETIME | orjson.OPT_SERIALIZE_NUMPY),
+    )
```
amotl, line 347: Let's break this out into a separate function.

amotl: Refactoring applied per a6b369e1be2.

src/crate/client/http.py
amotl: Because @seut says it works like [this](https://github.com/crate/crate-python/pull/691#pullrequestreview-2555959462), we are eating the free lunch now, and are able to remove quite an amount of code.

Thanks, @seut!

tests/client/test_http.py
```diff
@@ -724,10 +724,10 @@ def test_username(self):
...
-        self.assertEqual(result, "1687771440123")
+        result = json_dumps(data)
+        self.assertEqual(result, b'"2023-06-26T09:24:00.123000"')
 
     def test_aware_datetime(self):
         data = dt.datetime.fromisoformat("2023-06-26T09:24:00.123+02:00")
-        result = json.dumps(data, cls=CrateJsonEncoder)
-        self.assertEqual(result, "1687764240123")
+        result = json_dumps(data)
+        self.assertEqual(result, b'"2023-06-26T09:24:00.123000+02:00"')
```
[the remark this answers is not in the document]
amotl: Please review this patch once more. Thanks!

kneth (approved):
Great to see this performance fix. Thanks to @amotl and @seut!

seut (approved):
looks good

src/crate/client/http.py
```diff
@@ -86,25 +83,23 @@ def super_len(o):
...
+        return str(obj)
+    return obj
+
+
+def json_dumps(obj):
+    return orjson.dumps(
+        obj,
+        default=cratedb_json_encoder,
+        option=orjson.OPT_SERIALIZE_NUMPY,
+    )
```
amotl, line 102: ### Problem
Apparently, the current form of the marshaller lacks support for [freezegun](https://pypi.org/project/freezegun/) datetime fake types, i.e. `FakeDatetime`. It happens on the [CI run of a downstream project](https://github.com/crate/cloud-api/actions/runs/12810343874/job/35717190132?pr=3038#step:6:30891) per https://github.com/crate/cloud-api/pull/3038, where those are used within the test suite of the project.

```python
TypeError: default serializer exceeds recursion limit
```

### References
- https://github.com/spulec/freezegun/issues/448
- https://github.com/ijl/orjson/issues/481
- https://github.com/ijl/orjson/issues/245
- https://github.com/ijl/orjson/issues/37

amotl: This marshaller accounts for two details:

a) It supports freezegun's `FakeDatetime` objects.
b) It no longer raises `TypeError: default serializer exceeds recursion limit`, but `TypeError: Type is not JSON serializable: FakeDatetime` instead, which gives the user a better idea about what goes south, and why.

```python
from decimal import Decimal
from freezegun.api import FakeDatetime

if isinstance(obj, Decimal):
    return str(obj)
elif isinstance(obj, FakeDatetime):
    return obj.isoformat()
raise TypeError
```

amotl: In order not to pull in freezegun as a runtime dependency, the code will probably need to be phrased like this:

```python
from decimal import Decimal

if isinstance(obj, Decimal):
    return str(obj)
elif hasattr(obj, "isoformat"):
    return obj.isoformat()
raise TypeError
```

Is it the right way to do a duck typing match in Python, @surister, @juanpardo, @tomach?

surister:
Do we have a benchmark?

src/crate/client/http.py
surister: I don't have an issue against this, in opposition of a match statement. :+1:

src/crate/client/http.py
```diff
@@ -86,25 +83,23 @@ def super_len(o):
...
-            else:
-                delta = o - self.epoch_naive
-            return int(
-                delta.microseconds / 1000.0
-                + (delta.seconds + delta.days * 24 * 3600) * 1000.0
-            )
-        if isinstance(o, date):
-            return calendar.timegm(o.timetuple()) * 1000
-        return json.JSONEncoder.default(self, o)
+def cratedb_json_encoder(obj):
```
surister, line 86: NT: Type hints?

surister (approved):

amotl:
> Do we have a benchmark?

It's not a benchmark, but @widmogrod shared a flamegraph before/after over at https://github.com/crate/crate-python/issues/689#issuecomment-2595238024.

src/crate/client/http.py
amotl: Si. I've just amended the patch. Thanks!

amotl:
A new release `crate==2.0.0.dev4` has been published.

src/crate/client/http.py
amotl: A new release `crate==2.0.0.dev4` has been published, including this update.

src/crate/client/http.py
```diff
@@ -86,25 +84,33 @@ def super_len(o):
...
+    """
+    Serialize to JSON format, using `orjson`, with additional type support.
+
+    https://github.com/ijl/orjson
+    """
+    return orjson.dumps(
+        obj,
+        default=cratedb_json_encoder,
+        option=orjson.OPT_SERIALIZE_NUMPY,
+    )
```
amotl, line 113: ### Problem
On the [CI run of another downstream project](https://github.com/crate/cloud-api/actions/runs/12816234211/job/35736688530?pr=3038#step:6:3676):
```python
TypeError: Dict key must be str
```

### Solution
orjson can do it, it just needs to be configured correctly.

-- https://github.com/ijl/orjson#opt_non_str_keys

amotl: Done.
```python
return orjson.dumps(
    obj,
    default=cratedb_json_encoder,
    option=(orjson.OPT_NON_STR_KEYS | orjson.OPT_SERIALIZE_NUMPY),
)
```

PR #694, opened 2025-01-23
Remove tox, recommend uv

amotl:
## About
The configuration for [tox](https://pypi.org/project/tox/) apparently does not receive much maintenance, so let's remove it. Personally, I stopped using it since ages.

## Rationale
`tox` has been popular in times when it was easier to run a full Python test matrix on your workstation, for example supported by `pyenv`, compared to provisioning and maintaining it on a Jenkins CI runner.

Nowadays, it became so easy to run the full-version test matrix on GitHub Actions (GHA), so `tox` became obsolete in such environments that heavily use GHA. Now that recent advancements added `uv` to the table, sandbox environments don't need to feel ousted.

## References
- https://github.com/crate/crash/pull/461


seut (approved):
sound good.

PR #711, opened 2025-05-06
Fix: Fail early when database cluster does not respond

amotl:
## Problem

@shraik started using sqlalchemy-cratedb and reported that its behaviour deviates from other vendors by not failing on `engine.connect()` when the database server is not available.

- https://github.com/crate/sqlalchemy-cratedb/issues/218

We found this is not actually on the SQLAlchemy dialect, but on the DBAPI driver already, which exhibits the same behaviour.

## Solution

The patch extends the `_lowest_server_version` method to re-raise ~the last `ConnectionError` when no connection can be made to any configured server node~ a `ConnectionError` when connecting to all server nodes fails, including all error messages.

By doing it this way, we didn't need to submit a dummy SQL command like originally planned. We think it is much better this way because it does not pollute the server-side statement log.

## Details
As an additional benefit, the software tests in `test_connection.py` are now actual integration tests.

/cc @mfussenegger, @seut, @surister

### DIFF
--- a/CHANGES.rst
+++ b/CHANGES.rst
@@ -5,6 +5,8 @@ Changes for crate
 Unreleased
 ==========
 
+- Changed connection behaviour to fail early if the database cluster does not respond
+
 2026/03/09 2.1.2
 ================
 
--- a/CHANGES.rst
+++ b/CHANGES.rst
@@ -2,6 +2,13 @@
 Changes for crate
 =================
 
+Unreleased
+================
+
+- Breaking change: ``connect()`` now raises ``ConnectionError`` immediately if
+no configured server node responds. Changed connection behaviour to fail early 
+if the database cluster does not respond.
+
 2026/06/17 2.2.1
 ================
 
@@ -27,8 +34,6 @@ Changes for crate
   :class:`py:dict` as ``parameters`` to ``cursor.execute()`` now accepts
   ``%(name)s`` placeholders and converts them to positional ``?`` markers
   client-side. Positional parameters using ``?`` continue to work unchanged.
-
-- Changed connection behaviour to fail early if the database cluster does not respond
 
 2026/03/09 2.1.2
 ================
--- a/docs/by-example/client.rst
+++ b/docs/by-example/client.rst
@@ -29,12 +29,8 @@ respond, the request is automatically routed to the next server:
     >>> connection = client.connect([invalid_host, crate_host])
     >>> connection.close()
 
-If no ``servers`` are given, the default one ``http://127.0.0.1:4200`` is used:
-
-    >>> connection = client.connect()
-    >>> connection.client._active_servers
-    ['http://127.0.0.1:4200']
-    >>> connection.close()
+If no ``servers`` are supplied to the ``connect`` method, the default address
+``http://127.0.0.1:4200`` is used.
 
 If the option ``error_trace`` is set to ``True``, the client will print a whole
 traceback if a server error occurs:
--- a/docs/by-example/client.rst
+++ b/docs/by-example/client.rst
@@ -73,7 +73,7 @@ connect:
 
 The username for trusted users can also be provided in the URL:
 
-    >>> connection = client.connect(['http://trusted_me@' + crate_host])
+    >>> connection = client.connect([crate_host.replace('://', '://trusted_me@')])
     >>> connection.client.username
     'trusted_me'
     >>> connection.client.password
@@ -93,7 +93,7 @@ also need to provide ``password`` as argument for the ``connect()`` call:
 
 The authentication credentials can also be provided in the URL:
 
-    >>> connection = client.connect(['http://me:my_secret_pw@' + crate_host])
+    >>> connection = client.connect([crate_host.replace('://', '://me:my_secret_pw@')])
     >>> connection.client.username
     'me'
     >>> connection.client.password
--- a/src/crate/client/connection.py
+++ b/src/crate/client/connection.py
@@ -202,20 +202,21 @@ class Connection:
 
     def _lowest_server_version(self):
         lowest = None
-        last_connection_error = None
+        server_count = len(self.client.active_servers)
+        connection_errors = []
         for server in self.client.active_servers:
             try:
                 _, _, version = self.client.server_infos(server)
                 version = Version(version)
             except ConnectionError as ex:
-                last_connection_error = ex
+                connection_errors.append(ex)
                 continue
             except (ValueError, InvalidVersion):
                 continue
             if not lowest or version < lowest:
                 lowest = version
-        if lowest is None and last_connection_error is not None:
-            raise last_connection_error
+        if connection_errors and len(connection_errors) == server_count:
+            raise ConnectionError(str(connection_errors))
         return lowest or Version("0.0.0")
 
     def __repr__(self):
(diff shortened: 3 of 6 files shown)


src/crate/client/connection.py
```diff
@@ -197,14 +198,20 @@ def get_blob_container(self, container_name):
...
         lowest = None
+        last_connection_error = None
         for server in self.client.active_servers:
             try:
                 _, _, version = self.client.server_infos(server)
                 version = Version(version)
-            except (ValueError, ConnectionError):
+            except ConnectionError as ex:
+                last_connection_error = ex
+                continue
```
kneth, line 208: Won't this `continue` let you try all active servers? Do you want to stop at the first exception?

amotl: Yes and no.

docs/by-example/client.rst
```diff
@@ -35,12 +35,8 @@ respond, the request is automatically routed to the next server:
...
     >>> connection.close()
 
-If no ``servers`` are given, the default one ``http://127.0.0.1:4200`` is used:
-
-    >>> connection = client.connect()
-    >>> connection.client._active_servers
-    ['http://127.0.0.1:4200']
-    >>> connection.close()
+If no ``servers`` are supplied to the ``connect`` method, the default address
+``http://127.0.0.1:4200`` is used.
```
amotl, line 39: There is no server at `http://127.0.0.1:4200`. This test case just didn't fail because `connect()` did not raise an exception up until now.

Now, there is no longer a way to validate connecting to the default address per doctest file, because there is no server listening at `http://127.0.0.1:4200`. Because the core information is still viable, all what's left is pure prose, rephrased a bit.

src/crate/client/connection.py
kneth: I probably take the change log entry too literally.

CHANGES.rst
```diff
@@ -5,6 +5,9 @@ Changes for crate
 Unreleased
 ==========
 
+- Changed connection behaviour to fail early when database server
+  does not respond
```
amotl, line 9: @kneth: Maybe rephrase this statement like this?

```suggestion
- Changed connection behaviour to fail early when database cluster
  does not respond
```

surister (approved):
LGTM

src/crate/client/connection.py
matriv: Sorry, so what's happening? does it retry all servers provided and then throw the last connection error? (shouldn't it throw the first connection error maybe?)

seut: If it would throw the first error, the whole logic of supporting multiple servers would be invalidated I think.

seut (approved):
Lgtm

src/crate/client/connection.py
seut: ah sry got this wrong, not sure if throwing the first error would be better. both could make sense imho.

seut: another option could be to put all connection errors into an exception if all servers raise connection errors.

amotl: > If it would throw the first error, the whole logic of supporting multiple servers would be invalidated I think.

I share the same opinion that it would defeat the whole purpose, right.

amotl: @matriv: Do you like it better this way, now with 306feb5869791e?

src/crate/client/connection.py
```diff
@@ -197,14 +198,21 @@ def get_blob_container(self, container_name):
...
-            except (ValueError, ConnectionError):
+            except ConnectionError as ex:
+                connection_errors.append(ex)
+                continue
+            except (ValueError, InvalidVersion):
                 continue
             if not lowest or version < lowest:
                 lowest = version
+        if lowest is None and len(connection_errors) == server_count:
+            raise ConnectionError(str(connection_errors))
```
amotl, line 215: `str(connection_errors)` uses an opaque way, just serializing the list of exception instances into a string, the Python way. Do you have any objections or suggestions to do it differently?

amotl: > [...] suggestions to do it differently, like using JSON instead?

The procedure now uses JSON per e818221ac.

seut (approved):
👍

amotl:
We just published a pre-release package including those updates per [crate-2.1.0.dev0](https://pypi.org/project/crate/2.1.0.dev0/).

matriv (commented):
@amotl Thank you! Is there a way to test that we don't fail the connection attempt, but only if only all servers provided fail?
Can we also test the exception when multiple nodes fail?

src/crate/client/connection.py
```diff
@@ -197,14 +199,21 @@ def get_blob_container(self, container_name):
...
-            except (ValueError, ConnectionError):
+            except ConnectionError as ex:
+                connection_errors.append(ex)
+                continue
+            except (ValueError, InvalidVersion):
                 continue
             if not lowest or version < lowest:
                 lowest = version
+        if connection_errors and len(connection_errors) == server_count:
+            raise ConnectionError(json.dumps(list(map(str, connection_errors))))
```
surister, line 216: why not just `raise ConnectionError(error_list)`?

amotl: I've elaborated about it [here](https://github.com/crate/crate-python/pull/711#discussion_r2081480875), but I am also not sure, that's why I am also asking about your opinion.

kneth (approved):

amotl:
Hi @surister. The tests are failing now after rebasing on top of the recent test suite refactorings. Can I humbly ask you to look into it? Maybe you can spot the flaw quickly on the interface of both?

surister:
> Hi @surister. The tests are failing now after rebasing on top of the recent test suite refactorings. Can I humbly ask you to look into it? Maybe you can spot the flaw quickly on the interface of both?

looking

amotl:
Hi. I lost a bit track of this, maybe @surister or others can elaborate better. Could you also add tackling this stale patch to your backlog, @bgunebakan, when applicable?

bgunebakan:
@kneth @matriv @seut The PR was stale for a while, I updated the branch with main and fix did improvements on tests, log handling and docs. I kindly ask you to review again. Thanks in advance.

kneth (approved):

tests/client/test_connection.py
```diff
@@ -12,6 +13,40 @@
...
+    Create a Connection that bypasses __init__.
+    """
+
+    conn = Connection.__new__(Connection)
+    conn.client = client
+    return conn
+
+
+def test_invalid_server_address():
+    client = Client(servers="localhost:4202")
```
seut: Some other (manually started) node may listen on`4202`.
Let's use a complete different port to be safe ;) 

```suggestion
    client = Client(servers="localhost:1234")
```

CHANGES.rst
```diff
@@ -2,6 +2,13 @@
 Changes for crate
 =================
 
+Unreleased
+================
+
+- Breaking change: ``connect()`` now raises ``ConnectionError`` immediately if
+no configured server node responds. Changed connection behaviour to fail early 
+if the database cluster does not respond.
```
seut: Isn't the 2nd sentence repeating the first in some kind?

```suggestion
- Breaking change: ``connect()`` now raises ``ConnectionError`` immediately if
no configured server node responds.
```

seut (approved):
just some minor suggestions.

PR #717, opened 2025-05-24
Add JWT token authentication

amotl:
CrateDB 5.7.0 introduced JWT token authentication. Let's also implement it on the client sides.

- https://github.com/crate/crate/issues/14238


src/crate/client/http.py
```diff
@@ -189,6 +190,17 @@ def request(
             if length is not None:
                 headers["Content-Length"] = str(length)
 
+        # Sanity checks.
+        if jwt_token is not None and username is not None:
+            raise ValueError(
+                "Either JWT tokens are accepted, "
+                "or user credentials, but not both"
+            )
```
amotl: > Maybe adding a sanity check to give a good error if token and user/password auth is used in conjunction would be useful. 

Good idea. a1ac249a just added this additional sanity check as suggested.

PR #724, opened 2025-06-09
Remove python 3.7 and 3.8 from test matrix.

surister:
As you may know, both 3.7 and 3.8 are EOL, they receive no support/security fixes, nothing.

For that reason I would propose to remove them from the test matrix, we are also seeing new dependencies updates fail on 3.7, because of the libraries using >3.7 syntax (walrus operator), an example of this is https://github.com/crate/crate-python/actions/runs/15392143903/job/43303839083

We previously had a discussion on whether or not to keep 3.8 at https://github.com/crate/cratedb-sqlparse/pull/173

- Title

### DIFF
--- a/.github/workflows/tests.yml
+++ b/.github/workflows/tests.yml
@@ -19,7 +19,7 @@ jobs:
       fail-fast: false
       matrix:
         os: ['ubuntu-22.04']
-        python-version: ['3.7', '3.8', '3.9', '3.10', '3.11', '3.12', '3.13']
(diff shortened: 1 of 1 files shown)


amotl (approved):

amotl:
> We previously had a discussion on whether or not to keep 3.8 at https://github.com/crate/cratedb-sqlparse/pull/173.

To clarify, I would like to retain compatibility through package metadata, when applicable and given, even when dropping validation on CI, if it works and if it's reasonably responsible.

PR #737, opened 2025-10-10
Update Python version to include 3.14

surister:

Title.


surister:
Added 3.9 back since we have a lot of users still on 3.9

amotl (approved):

PR #755, opened 2025-12-18
Chore: Use SPDX (PEP 639). Update verlib2.

amotl:
Just maintenance.

### DIFF
--- a/setup.py
+++ b/setup.py
@@ -57,7 +57,7 @@ setup(
     install_requires=[
         "orjson<4",
         "urllib3",
-        "verlib2",
+        "verlib2>=0.3",
     ],
     extras_require={
         "doc": [


setup.py
amotl: I think `setup.py` can be removed soon?

pyproject.toml
```diff
@@ -15,11 +15,11 @@ description = "CrateDB Python Client"
 authors = [{ name = "Crate.io", email = "office@crate.io" }]
 requires-python = ">=3.10"
 readme = "README.rst"
-license = { file = "LICENSE"}
+license = "Apache-2.0"
+license_files = ["LICENSE"]
```
seut: According to https://packaging.python.org/en/latest/guides/licensing-examples-and-user-scenarios/ there is no need to explicitly define `license_files` or `license_file`?
```suggestion
```

seut (approved):
Isn't this more related to https://peps.python.org/pep-0639/? Afaik we used SPDX expressions even before. I'd suggest to update the commit message to reflect this.

PR #768, opened 2026-01-19
Re-add execution of doctests

mfussenegger:
Follow up to https://github.com/crate/crate-python/pull/767


docs/by-example/client.rst
```diff
@@ -25,7 +25,7 @@ replication. In order for clients to make use of this property it is
 recommended to specify all hosts of the cluster. This way if a server does not
 respond, the request is automatically routed to the next server:
 
-    >>> invalid_host = 'http://not_responding_host:4200'
+    >>> invalid_host = 'http://127.0.0.1:4201'
```
mfussenegger: Haven't investigated why, but on my machine resolving a hostname that doesn't exist takes way longer than having it resolve to localhost on a port where hopefully nothing runs.

Potential follow up would be to generally restructure this. http.rst and client.rst have lots of overlap.

tests/testing/tests.py
```diff
@@ -1,34 +0,0 @@
...
-# software solely pursuant to the terms of the relevant commercial agreement.
-
-import unittest
-
-from .test_layer import LayerTest, LayerUtilsTest
-
-makeSuite = unittest.TestLoader().loadTestsFromTestCase
-
-
-def test_suite():
```
mfussenegger: pytest afaik doesn't support the test_suite discovery. Tests are picked up based on filename pattern and LayerTest/LayerUtilsTest still get executed

tests/test_docs.py
```diff
@@ -0,0 +1,26 @@
...
+from .conftest import crate_host
+
+
+def cprint(s):
+    if isinstance(s, bytes):
+        s = s.decode("utf-8")
+    print(s)  # noqa: T201
+
+
+def test_docs(doctest_node):
```
mfussenegger: I also tried `--doctest-glob="*.rst"` (https://docs.pytest.org/en/stable/how-to/doctest.html) instead with a autouse fixture, but that would also be used in the other tests and we've have two crate instances running in parallel.

So this seemed like the better option.

PR #769, opened 2026-01-20
Remove py < 3.9 deps; Mention dropped python version support in changes

mfussenegger:
Also removes the "Modernize project definition" entry. CHANGES.rst
should be written for users of the library. The entry had no relevance
for end users.

### DIFF
--- a/pyproject.toml
+++ b/pyproject.toml
@@ -49,7 +49,6 @@ classifiers = [
         "Topic :: System :: Networking",
 ]
 dependencies = [
-    "importlib-metadata; python_version<'3.8'",
     "orjson",
     "urllib3",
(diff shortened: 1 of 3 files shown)


pyproject.toml
```diff
@@ -49,15 +49,13 @@ classifiers = [
...
 dependencies = [
-    "importlib-metadata; python_version<'3.8'",
     "orjson",
     "urllib3",
     "verlib2>=0.3.1",
 ]
 
 [dependency-groups]
 dev = [
-    "backports.zoneinfo<1; python_version<'3.9'",
```
amotl: Hi. This would contradict the other patch.

- GH-759

PR #802, opened 2026-05-15
Add Support HTTP compression

### LINKED ISSUE #700: Support HTTP compression
## About

@proddata said:

> It seems like CrateDB's HTTP interface accepts gzip / deflate compressed data.
> It might also be interesting to add this capability to `crate-python`.

## References

- https://github.com/crate/crate/pull/17494

bgunebakan:
Adds opt-in HTTP compression support to the crate-python driver to reduce bandwidth usage and improve performance for large queries.

- Client-side request compression (compress_client, default: True). Outgoing JSON request bodies are gzip-compressed before being sent.
- Server-side response compression (compress_server, default: False). The driver sends `Accept-Encoding: gzip, deflate` to CrateDB server when it's enabled.
- Compression is skipped for payloads below `compress_threshold` bytes (default 8192)
- 6 new unit tests covering: compression disabled, compression enabled, below-threshold skip, Accept-Encoding sent/not sent, and default-args behavior.

### Small note on threshold default value:
I've changed `compress_threshold` to 8192 while implementing. I ran a payload-size sweep against both a local CrateDB instance and a CrateDB Cloud cluster (eu-west-1 free tier), measuring round-trip latency with and without gzip compression. I saw benefits of compression when latency included over 8kb on local and 1kb on cloud instances. 

Latency can vary based on my network and cluster settings/location. I decided to move on 8kb as a default value for threshold. 

## Checklist

 - [X] Link to issue this PR refers to (if applicable): Fixes #700 
 - [X] Added or Changed code is covered by tests
 - [X] Updated documentation & CHANGES.rst

### DIFF
--- a/src/crate/client/http.py
+++ b/src/crate/client/http.py
@@ -464,7 +464,7 @@ class Client:
         socket_tcp_keepintvl=None,
         socket_tcp_keepcnt=None,
         jwt_token=None,
-        compress=8192,
+        compress: t.Union[int, bool] = 8192,
     ):
         if not servers:
             servers = [self.default_server]
@@ -489,7 +489,7 @@ class Client:
                 )
 
         self._active_servers = servers
-        self._inactive_servers = []
+        self._inactive_servers: t.List[t.Tuple[float, str, str]] = []
         pool_kw = _pool_kw_args(
             verify_ssl_cert,
             ca_cert,
@@ -508,7 +508,7 @@ class Client:
         )
         self.ssl_relax_minimum_version = ssl_relax_minimum_version
         self.backoff_factor = backoff_factor
-        self.server_pool = {}
+        self.server_pool: t.Dict[str, Server] = {}
         self._update_server_pool(servers, **pool_kw)
         self._pool_kw = pool_kw
         self._lock = threading.RLock()
@@ -517,6 +517,11 @@ class Client:
         self.password = password
         self.jwt_token = jwt_token
         self.schema = schema
+
+        if not isinstance(compress, (bool, int)):
+            raise TypeError(
+                f"compress must be bool or int, got {type(compress).__name__!r}"
+            )
         self.compress = compress
 
         self.path = self.SQL_PATH
--- a/src/crate/client/http.py
+++ b/src/crate/client/http.py
@@ -706,7 +706,9 @@ class Client:
             data = gzip.compress(data, compresslevel=6)
             headers["Content-Encoding"] = "gzip"
 
-        response = self._request(method, path, data=data, headers=headers or None)
+        response = self._request(
+            method, path, data=data, headers=headers or None
+        )
         _raise_for_status(response)
         if len(response.data) > 0:
             return _json_from_response(response)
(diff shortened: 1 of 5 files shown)


mfussenegger (commented):
- I think we can remove the `compress_server` setting and always set the header leaving it up to the server - given that in CrateDB compression also needs to be enabled with default being off.
- I'd also remove the `compress_algorithm` setting - given that there is so far only gzip supported there's no point in the setting
- I'd tend to merge `compress_client` / `compress_threshold` to a single setting:
    - number: Compressed if payload body > the number
    - bool true: Always compressed
    - bool false: Never compressed
  Otherwise the interaction between enabled/threshold seems more ambiguous and also sort of redundant to have both separate.

Didn't look into the details yet

bgunebakan:
Thanks for your review!

> * I think we can remove the `compress_server` setting and always set the header leaving it up to the server - given that in CrateDB compression also needs to be enabled with default being off.

Agree to remove it and always send `Accept-Encoding: gzip, deflate`. I had it as an opt-in partly out of caution around BREACH, but we can leave this control to server config.

> * I'd also remove the `compress_algorithm` setting - given that there is so far only gzip supported there's no point in the setting

Agree to remove it if there is no plan to implement other algorithms.

> * I'd tend to merge `compress_client` / `compress_threshold` to a single setting:
>   
>   * number: Compressed if payload body > the number
>   * bool true: Always compressed
>   * bool false: Never compressed
>     Otherwise the interaction between enabled/threshold seems more ambiguous and also sort of redundant to have both separate.
> 
> Didn't look into the details yet

Make sense to merge compress parameters. The type check needs to test for bool before int, otherwise `compress=True` would be interpreted as threshold `1`.

src/crate/client/http.py
```diff
@@ -463,6 +464,7 @@ def __init__(
         socket_tcp_keepintvl=None,
         socket_tcp_keepcnt=None,
         jwt_token=None,
+        compress=8192,
```
mfussenegger:
```suggestion
        compress: Union[int, bool] = 8192,
```

I think it would be good if we started adding type annotations - at least for the public API

src/crate/client/http.py
```diff
@@ -678,8 +681,16 @@ def _json_request(self, method, path, data):
         """
         Issue request against the crate HTTP API.
         """
+        headers = {"Accept-Encoding": "gzip, deflate"}
 
-        response = self._request(method, path, data=data)
+        compress_enabled = self.compress is True or (
+            not isinstance(self.compress, bool) and len(data) >= self.compress
+        )
```
mfussenegger, line 693:
```suggestion
        compress_enabled = self.compress is True or (
            isinstance(self.compress, int) and len(data) >= self.compress
        )
```

Or might even make it stricter and fail if it's neither bool nor int.

mfussenegger (approved):
Left some minor suggestions, otherwise lgtm

src/crate/client/http.py
bgunebakan: I added types, I'd like to tackle all annotations in separate PR. I needed to add for `_inactive_servers` and `server_pool` because mypy were complaining.

src/crate/client/http.py
bgunebakan: I added check in `__init__` and will fail if it's not `int` or `bool`

PR #810, opened 2026-06-08
Add support for executemany()

### LINKED ISSUE #807: `executemany()` does not translate pyformat dict params
### Summary

`Cursor.executemany()` with pyformat `%(name)s` SQL and a list of dicts raises a `SQLParseException` in crate-python 2.2.0. The pyformat→positional translation introduced in dev is never applied on the bulk code path, raw `%(name)s` SQL is forwarded to CrateDB's HTTP API, which does not understand pyformat syntax.

### Root cause

`executemany()` **does not** passes `seq_of_parameters` as `parameters`:

```python
# src/crate/client/cursor.py, line 113–121
def executemany(self, sql, seq_of_parameters):
    ...
    self.execute(sql, bulk_parameters=seq_of_parameters)  # ← bulk path
```

`execute()` only calls `_convert_named_to_positional()` when `isinstance(parameters, dict)` (line 101). The `bulk_parameters` argument is never inspected for pyformat translation.

As a result, the following JSON body is sent to CrateDB's `/_sql` endpoint:

```json
{
  "stmt": "INSERT INTO t (name, age) VALUES (%(name)s, %(age)s)",
  "bulk_args": [{"name": "Arthur", "age": 42}]
}
```

CrateDB's SQL parser rejects the `%(...)s` syntax:

```
SQLParseException[Failed to parse source [...]]
```

### Reproduction

```python
import crate.client as crate_client

conn = crate_client.connect("localhost:4200")
cur = conn.cursor()
cur.execute(
    "CREATE TABLE IF NOT EXISTS t (name TEXT PRIMARY KEY, age INT) "
    "WITH (number_of_replicas = 0)"
)

# FAILS
cur.executemany(
    "INSERT INTO t (name, age) VALUES (%(name)s, %(age)s)",
    [
        {"name": "Arthur", "age": 42},
        {"name": "Trillian", "age": 35},
    ],
)
```

### Related Issues: 
- Issue #806 
- PR #795

bgunebakan:
## Checklist

 - [x] Link to issue this PR refers to (if applicable): Fixes #807

### DIFF
--- a/src/crate/client/cursor.py
+++ b/src/crate/client/cursor.py
@@ -70,6 +70,37 @@ def _convert_named_to_positional(
     return converted_sql, new_params
 
 
+def _convert_named_bulk_params(
+    sql: str, seq_of_dicts: t.Sequence[t.Dict[str, t.Any]]
+) -> t.Tuple[str, t.List[t.List[t.Any]]]:
+    """Convert pyformat SQL and a sequence of dicts to positional bulk args.
+
+    Uses the first row to determine the SQL template and position map, then
+    builds a positional argument list for every row.
+
+    Raises ``ProgrammingError`` if a placeholder name is absent from any row.
+    Extra keys in each row are silently ignored (consistent with
+    ``_convert_named_to_positional``).
+    """
+    first = seq_of_dicts[0]
+    converted_sql, _ = _convert_named_to_positional(sql, first)
+    positions = {k: i + 1 for i, k in enumerate(first)}
+    n = len(positions)
+
+    bulk_args: t.List[t.List[t.Any]] = []
+    for row in seq_of_dicts:
+        positional: t.List[t.Any] = [None] * n
+        for name, pos in positions.items():
+            if name not in row:
+                raise ProgrammingError(
+                    f"Named parameter '{name}' not found in the parameters dict"
+                )
+            positional[pos - 1] = row[name]
+        bulk_args.append(positional)
+
+    return converted_sql, bulk_args
+
+
 class Cursor:
     """
     not thread-safe by intention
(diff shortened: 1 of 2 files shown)


kneth (approved):
LGTM

src/crate/client/cursor.py
```diff
@@ -118,7 +149,16 @@ def executemany(self, sql, seq_of_parameters):
         """
         row_counts = []
         durations = []
-        self.execute(sql, bulk_parameters=seq_of_parameters)
+        bulk_parameters = seq_of_parameters
+        if (
+            bulk_parameters
+            and isinstance(bulk_parameters[0], dict)
+            and _NAMED_PARAM_RE.search(sql)
```
mfussenegger, line 161: What happens if there is a mix of dict/non-dict args?

tests/client/test_cursor.py
```diff
@@ -125,6 +125,53 @@ def test_cursor_executemany(mocked_connection):
...
+        "cols": [],
+        "duration": 123,
+        "results": [{"rowcount": 1}, {"rowcount": 1}],
+    }
+    with mock.patch.object(
+        mocked_connection.client, "sql", return_value=response
+    ):
+        cursor = mocked_connection.cursor()
+        cursor.executemany(
+            "INSERT INTO characters (name, age) VALUES (%(name)s, %(age)s)",
```
mfussenegger, line 145: Can you also add a case where one named argument is used multiple times and check that it gets the same position number?

src/crate/client/cursor.py
bgunebakan: currently we raise an error. Do we have case to use mix of dict/non-dict together? I couldn't find how server handle that. What it should do?

mfussenegger: Raising an error sounds good if it's one that makes it clear that mixing dict/non-dict together isn't supported.

bgunebakan: I added edge case [here cursor.py:92](https://github.com/crate/crate-python/pull/810/changes#diff-fd97e585850a385a9eca471196fdfc76201e7cb13ac02fd3cd507d249d80d2f5R92) so it will raise `ProgrammingError` and added test for that.

tests/client/test_cursor.py
bgunebakan: Added tests https://github.com/crate/crate-python/pull/810/changes/10faeae4b4558652c9d8f83e3ef7c03f088ba226

mfussenegger (approved):

PR #812, opened 2026-06-10
Add support for converting pyformat SQL to positional markers with bulk_parameters

bgunebakan:
Fixes a silent runtime error when `cursor.execute()` is called with `bulk_parameters` and a SQL statement that uses `%(name)s` (pyformat) placeholders.

When `bulk_parameters` is passed alongside pyformat SQL, the driver was sending the raw `%(name)s` markers verbatim to CrateDB, which rejected them with:

`SQLParseException[no viable alternative at input 'VALUES (%']`

The fix adds an elif branch in `execute()` that rewrites the SQL template to $N positional markers before the request is sent.

The bug was introduced indirectly by PR #810 which declared `paramstyle = "pyformat"` for the driver. Before that change, SQLAlchemy generated `?` style SQL, so the `bulk_parameters` path was never fed pyformat placeholders so the broken code path simply didn't exist.

It fixes implementation in sqlalchemy-crate package crate/sqlalchemy-cratedb#266

## Checklist

 - [ ] Link to issue this PR refers to (if applicable): Fixes #???
 - [x] Unit tests added

### DIFF
--- a/docs/by-example/cursor.rst
+++ b/docs/by-example/cursor.rst
@@ -266,6 +266,35 @@ For completeness' sake the cursor description is updated nonetheless:
     >>> [ desc[0] for desc in cursor.description ]
     ['name', 'position']
 
+executemany() with named parameters
+====================================
+
+``executemany()`` also accepts a :class:`py:list` of :class:`py:dict` when
+the SQL uses ``%(name)s`` placeholders. The client converts both the SQL
+template and all rows to positional format before sending to CrateDB:
+
+    >>> connection.client.set_next_response({
+    ...     "results": [
+    ...         {"rowcount": 1},
+    ...         {"rowcount": 1}
+    ...     ],
+    ...     "duration": 123,
+    ...     "cols": [],
+    ... })
+
+    >>> cursor = connection.cursor()
+
+    >>> cursor.executemany(
+    ...     "INSERT INTO t (id, val) VALUES (%(id)s, %(val)s)",
+    ...     [{"id": 1, "val": "foo"}, {"id": 2, "val": "bar"}])
+    [{'rowcount': 1}, {'rowcount': 1}]
+
+    >>> cursor.rowcount
+    2
+
+    >>> cursor.duration
+    123
+
     >>> connection.client.set_next_response({
     ...     "rows":[ [ "North West Ripple", 1 ], [ "Arkintoofle Minor", 3 ], [ "Alpha Centauri", 3 ] ],
     ...     "cols":[ "name", "position" ],
--- a/docs/by-example/cursor.rst
+++ b/docs/by-example/cursor.rst
@@ -270,8 +270,8 @@ executemany() with named parameters
 ====================================
 
 ``executemany()`` also accepts a :class:`py:list` of :class:`py:dict` when
-the SQL uses ``%(name)s`` placeholders. The client converts both the SQL
-template and all rows to positional format before sending to CrateDB:
+the SQL statement contains ``%(name)s`` placeholders. The client converts both the SQL
+template and all rows to positional format before sending them to CrateDB:
 
     >>> connection.client.set_next_response({
     ...     "results": [
--- a/docs/by-example/cursor.rst
+++ b/docs/by-example/cursor.rst
@@ -273,6 +273,8 @@ executemany() with named parameters
 the SQL statement contains ``%(name)s`` placeholders. The client converts both the SQL
 template and all rows to positional format before sending them to CrateDB:
 
+.. Hidden: set up mocked response
+
     >>> connection.client.set_next_response({
     ...     "results": [
     ...         {"rowcount": 1},
--- a/docs/query.rst
+++ b/docs/query.rst
@@ -72,20 +72,20 @@ The same parameter name may appear multiple times in the query:
     ...     "SELECT * FROM locations WHERE name = %(q)s OR kind = %(q)s",
     ...     {"q": "Quasar"})
 
-The client converts the ``%(name)s`` placeholders to positional ``?`` markers
-before sending the query to CrateDB, so no server-side changes are required.
-
-.. NOTE::
-
-    Named parameters are not yet supported by ``executemany()``. Use
-    positional ``?`` placeholders with a :class:`py:list` of tuples for bulk
-    operations.
+The client converts the ``%(name)s`` placeholders to ``$N`` positional
+markers before sending the query to CrateDB, so no server-side changes are
+required.
 
 Bulk inserts
 ------------
 
 :ref:`Bulk inserts <crate-reference:http-bulk-ops>` are possible with the
-``executemany()`` method, which takes a :class:`py:list` of tuples to insert:
+``executemany()`` method.
+
+Positional parameters
+.....................
+
+Pass a :class:`py:list` of tuples using ``?`` placeholders:
 
     >>> cursor.executemany(
     ...     "INSERT INTO locations (name, date, kind, position) VALUES (?, ?, ?, ?)",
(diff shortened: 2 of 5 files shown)


src/crate/client/cursor.py
```diff
@@ -136,6 +143,8 @@ def execute(self, sql, parameters=None, bulk_parameters=None):
 
         if isinstance(parameters, dict):
             sql, parameters = _convert_named_to_positional(sql, parameters)
+        elif bulk_parameters is not None and _NAMED_PARAM_RE.search(sql):
+            sql = _rewrite_pyformat_sql(sql)
```
mfussenegger: Wouldn't the bulk_parameters also need some sort of conversion and be in dict format for this to work reliably?

Or does pyformat imply that the bulk parameters order matches the order of named placeholders? If so - is that documented somewhere?

Do other clients allow this?

bgunebakan: > Wouldn't the bulk_parameters also need some sort of conversion and be in dict format for this to work reliably?

I did more tests and you're right that the original implementation was fragile so I added new check for dict format.
> 
> Or does pyformat imply that the bulk parameters order matches the order of named placeholders? If so - is that documented somewhere?
> 
This isn't specified in DB-API 2.0, `bulk_parameters` is a CrateDB specific and not covered by the spec.
> Do other clients allow this?

Other clients does not use this, psycopg3 don't expose a `bulk_parameters` parameter, it uses `executemany()` or `COPY` methods.

I'm extending the docs to cover this.

docs/by-example/cursor.rst
```diff
@@ -266,6 +266,35 @@ For completeness' sake the cursor description is updated nonetheless:
     >>> [ desc[0] for desc in cursor.description ]
     ['name', 'position']
 
+executemany() with named parameters
+====================================
+
+``executemany()`` also accepts a :class:`py:list` of :class:`py:dict` when
+the SQL uses ``%(name)s`` placeholders. The client converts both the SQL
```
mfussenegger:
```suggestion
``executemany()`` also accepts a :class:`py:list` of :class:`py:dict` when
the SQL statement contains ``%(name)s`` placeholders. The client converts both the SQL
```

docs/by-example/cursor.rst
```diff
@@ -266,6 +266,35 @@ For completeness' sake the cursor description is updated nonetheless:
...
+template and all rows to positional format before sending to CrateDB:
```
mfussenegger:
```suggestion
template and all rows to positional format before sending them to CrateDB:
```

docs/by-example/cursor.rst
```diff
@@ -266,6 +266,35 @@ For completeness' sake the cursor description is updated nonetheless:
...
+
+    >>> connection.client.set_next_response({
+    ...     "results": [
+    ...         {"rowcount": 1},
+    ...         {"rowcount": 1}
+    ...     ],
+    ...     "duration": 123,
+    ...     "cols": [],
+    ... })
```
mfussenegger, line 285: Shouldn't this be hidden from the public docs given that it's mocking the response and isn't a public API that should/can be used?

docs/query.rst
```diff
@@ -72,20 +72,20 @@ The same parameter name may appear multiple times in the query:
...
-before sending the query to CrateDB, so no server-side changes are required.
-
-.. NOTE::
-
-    Named parameters are not yet supported by ``executemany()``. Use
-    positional ``?`` placeholders with a :class:`py:list` of tuples for bulk
-    operations.
+The client converts the ``%(name)s`` placeholders to ``$N`` positional
+markers before sending the query to CrateDB, so no server-side changes are
+required.
```
mfussenegger:
```suggestion
The client converts the ``%(name)s`` placeholders to ``$N`` positional
markers before sending the query to CrateDB.
```

That remark probably isn't all that relevant for users. I doubt anyone would want to patch CrateDB to make things work and it is kind of implying that would be an option.

docs/query.rst
```diff
@@ -72,20 +72,20 @@ The same parameter name may appear multiple times in the query:
...
 ------------
 
 :ref:`Bulk inserts <crate-reference:http-bulk-ops>` are possible with the
-``executemany()`` method, which takes a :class:`py:list` of tuples to insert:
+``executemany()`` method.
+
+Positional parameters
+.....................
+
+Pass a :class:`py:list` of tuples using ``?`` placeholders:
```
mfussenegger: I think list or other type of sequences work too, not only tuples?

docs/query.rst
```diff
@@ -94,10 +94,56 @@ Bulk inserts
...
+    ...     [{"name": "Cloverleaf", "date": "2007-03-11", "kind": "Quasar", "pos": 7},
+    ...      {"name": "Old Faithful", "date": "2007-03-11", "kind": "Quasar", "pos": 8}])
+    [{'rowcount': 1}, {'rowcount': 1}]
+
+Using ``bulk_parameters`` directly
+...................................
+
+For advanced use cases (such as SQLAlchemy integrations) you can call
+``execute()`` directly with the ``bulk_parameters`` keyword argument,
+bypassing ``executemany()``:
```
mfussenegger: Just curios: Why is this relevant for SQLAlchemy?

To me it is generally not clear what advantage this offers over using `executemany`?

Judging from https://github.com/crate/crate-python/commit/d4223cd2a5db99dee9db29b2bdb4703beff182c6 it appears to me as if this was more of a side effect of re-using some logic.

docs/query.rst
bgunebakan: Ah you're right, any sequence works, I'm changing it to sequences.
```suggestion
Pass a :class:`py:list` of sequences using ``?`` placeholders:
```

docs/by-example/cursor.rst
```diff
@@ -266,6 +266,35 @@ For completeness' sake the cursor description is updated nonetheless:
...
     ['name', 'position']
 
+executemany() with named parameters
+====================================
+
+``executemany()`` also accepts a :class:`py:list` of :class:`py:dict` when
+the SQL statement contains ``%(name)s`` placeholders. The client converts both the SQL
+template and all rows to positional format before sending them to CrateDB:
+
+    >>> connection.client.set_next_response({
```
bgunebakan, line 278:
```suggestion
.. Hidden: set up mocked response

    >>> connection.client.set_next_response({
```

docs/query.rst
bgunebakan: There is no server level performance improvements compared to `executemany()`. I didn't do any performance tests to both methods but they use same http level, just `executemany()` does iteration on python level so they could be a difference based on data size. probably relatively small difference.

Do you think should we remove "Using ``bulk_parameters`` directly" part from public docs?

mfussenegger: > Do you think should we remove "Using bulk_parameters directly" part from public docs?

Given the type signature is public API I think it is fine to keep some docs, but I'd tend to nudge users to `executemany()` because that's DB API and merely mention that `execute(..., bulk_parameters)` exists as an alternative.

bgunebakan:
```suggestion
``execute()``  accepts a ``bulk_parameters`` keyword argument directly:

.. NOTE::
    Please prefer ``executemany()`` for bulk inserts, it is the standard DB-API 2.0
    interface. The ``bulk_parameters`` argument is a lower-level alternative.
```
maybe note like that?

docs/by-example/cursor.rst
bgunebakan: I added hidden keyword.

docs/query.rst
mfussenegger: Sounds good

mfussenegger (approved):

PR #828, opened 2026-09-02
Fix Register converter for DataType.BIT

### LINKED ISSUE #817: Register converter for DataType.BIT
`DataType.BIT = 25` is defined in `converter.py`, but no converter function is registered for it in `_DEFAULT_CONVERTERS`. BIT values therefore fall through `_to_default` unchanged and reach Python as CrateDB's `B'…'` SQL literal form, which cannot be fed back to the server as a parameter, making the read/write round-trip asymmetric.

## Current behaviour

```sql
CREATE TABLE t_bit (id INT, flags BIT(8));
```

```python
cur.execute("INSERT INTO t_bit (id, flags) VALUES (1, ?)", ("00000001",))  # bare bits
cur.execute("SELECT flags FROM t_bit WHERE id = 1")
value = cur.fetchone()[0]
print(value)   # -> "B'00000001'"   (literal wrapper, not what was written)
```

The value that comes out is not accepted going back in:

```python
cur.execute("INSERT INTO t_bit (id, flags) VALUES (2, ?)", (value,))
# ProgrammingError: SQLParseException[Cannot cast value `B'00000001'` to type `bit(8)`]

cur.execute("SELECT count(*) FROM t_bit WHERE flags = ?", (value,))
# ProgrammingError: SQLParseException[Cannot cast value `B'00000001'` to type `bit(8)`]
```
## Proposed change

Register a `_to_bit_string` converter for `DataType.BIT` that strips the literal wrapper and returns the bare bit string.

bgunebakan:
`DataType.BIT` had an enum member but no entry in `_DEFAULT_CONVERTERS`, so `BIT` values fell through `_to_default` and reached Python in CrateDB's SQL literal form, `B'0110'`. That form is not accepted back as a query parameter, which made the round-trip asymmetric.

```
cur.execute("SELECT flags FROM t")            # -> "B'00000001'"
cur.execute("... WHERE flags = ?", (value,))  # ProgrammingError: Cannot cast value
_to_bit_string now strips the literal wrapper and returns a plain string of 0/1 digits
```

`_to_bit_string` now strips the literal wrapper and returns a plain string of `0/1` digits, so what you read is what you can write.

## Checklist

 - [x] Link to issue this PR refers to (if applicable): Fixes #817 
 - [x] Increased test coverage
 - [x] Added required documentation


docs/by-example/cursor.rst
```diff
@@ -379,6 +380,25 @@ and decoded to a ``datetime.time`` object with the appropriate timezone:
...
+    ...     "rows":[ [ "B'0110'" ] ],
+    ...     "cols":[ "flags" ],
+    ...     "rowcount":1,
+    ...     "duration":1
+    ... })
+
+    >>> cursor.execute('')
+
+    >>> cursor.fetchone()
+    ['0110']
```
mfussenegger, line 403: This looks rather odd as documentation - especially the `cursor.execute('')` will likely be rather confusing.  Can we hide the `set_next_response` and have a `cursor.execute("select b'0110'")` instead?

src/crate/client/converter.py
```diff
@@ -26,13 +26,16 @@
...
 import ipaddress
+import re
 from copy import deepcopy
 from enum import Enum
 from typing import Any, Callable, Dict, List, Optional, Union
 
 ConverterFunction = Callable[[Optional[Any]], Optional[Any]]
 ColTypesDefinition = Union[int, List[Union[int, "ColTypesDefinition"]]]
 
+_BIT_LITERAL = re.compile(r"B'([01]*)'")
```
mfussenegger, line 37:
```suggestion
_BIT_LITERAL = re.compile(r"B'([01]+)'")
```

I think empty bitstrings are not possible? But please double check

mfussenegger (approved):
Left two suggestions - otherwise lgtm

src/crate/client/converter.py
florinutz: hey, I checked, they work:
```
>>> SELECT B''
{"cols":["B''"],"col_types":[25],"rows":[["B''"]],"rowcount":1}
```

```
>>> SELECT ''::bit(0)
{"cols":["CAST('' AS bit(0))"],"col_types":[25],"rows":[["B''"]],"rowcount":1}
```

```
>>> SELECT B'0110'::bit(0)
{"cols":["CAST(B'0110' AS bit(0))"],"col_types":[25],"rows":[["B''"]],"rowcount":1}
```

```
>>> CREATE TABLE tabel (b bit(0))
{"cols":[],"col_types":[],"rows":[[]],"rowcount":1}
```

---

and unrelated to this ticket:

```
>>> SELECT ''::bit(1)
{"error":{"message":"StringIndexOutOfBoundsException[Index 0 out of bounds for length 0]","code":5000}}
```

```
>>> SELECT NULL::bit(4)
{"error":{"message":"NullPointerException[Cannot invoke \"io.crate.sql.tree.BitString.length()\" because \"bs\" is null]","code":5000}}
```

florinutz (approved):
looks good!

the 2 statements I found return http 500 with a java exception instead of a SQL error (crate 6.4.1). They are out of the scope of this ticket, and I think we should raise it as bug(s?) for crate/crate

docs/by-example/cursor.rst
bgunebakan: Thanks, you are right. I changed it to `cursor.execute("select b'0110'")`. I didn't understand what you meant "hide the `set_next_response`"

mfussenegger: `hide:` directives like we use in the CrateDB docs. E.g.:

https://github.com/crate/crate/blob/8900fd40976c994d14a2097daff59ea32af3bff8/docs/general/information-schema.rst?plain=1#L50-L50

This way it won't show up in the rendered documentation.

bgunebakan:
> looks good!
> 
> the 2 statements I found return http 500 with a java exception instead of a SQL error (crate 6.4.1). They are out of the scope of this ticket, and I think we should raise it as bug(s?) for crate/crate

I also reproduced, and yes, looks like a bug. I think crate's 500 codes shouldn't reach all the way here, they should report as a 4000 error. Could you create a bug report on crate/crate?

src/crate/client/converter.py
mfussenegger: Both will be fixed in 6.4.5 with https://github.com/crate/crate/pull/20155 and https://github.com/crate/crate/pull/20156

PR #830, opened 2026-09-10
Fix ValueError on unlisted CrateDB column type identifiers

### LINKED ISSUE #829: DataType enum: reads raise ValueError on five documented wire type IDs
## Summary & Why It is important

`Converter.get()` resolves `DataType(type_)` before it can fall back to the default converter, so any column type identifier the enum doesn't list raises instead of passing the value through. Five documented IDs are missing: 
1. `17 INTERVAL`
2. `18 ROW`
3. `28 FLOAT_VECTOR`
4. `29 UUID`
5. `30 REGTYPE`.

## Impact
`FLOAT_VECTOR` is the sharp edge. Critically, users who never touch converters still hit this, setting `cursor.time_zone` implicitly constructs a Converter, so a timezone-aware cursor breaks on a vector column. A plain cursor is unaffected, which makes it look config-dependent to users.

## Reproduction
Verified against a live `CrateDB 6.4.3`. I derived all five IDs from the server rather than trusting the docs table:

ID	| Type	| Statement|	Wire value |
-----|--------|-----------|--------------|
17	| INTERVAL |	SELECT INTERVAL '1 day' |	"1 day 00:00:00" |
18	| ROW	| SELECT pg_get_keywords() LIMIT 1	| ["over","U","unreserved"] |
28	| FLOAT_VECTOR |	SELECT embedding FROM t |	[0.1, 0.2] |
29	| UUID |	SELECT '…'::uuid |	"a5b3c1e0-…" |
30	| REGTYPE |	SELECT 23::regtype |	"int4" |

All five fail on both the explicit-converter path and the `time_zone=`-only path. The plain-cursor control returns them fine.

## Root cause
`DataType(type_)` is constructed to be used as a dict key, so an unlisted ID raises before `self._default` is reachable.

bgunebakan:
- Added the five missing members to `DataType`. 
- Added `_resolve()` that returns `None` for identifiers unknown to this client instead of raising error.
- Add UUID support to DefaultTypeConverter and update documentation

It is out of the scope of the issue ticket but I added a converter for `DataType.UUID` in this PR as well, because the enum member is introduced here.Before the round trip was asymmetric. `uuid.UUID` is already accepted as a parameter on the write side, but reads came back as `str`. `DefaultTypeConverter` now decodes UUID columns to `uuid.UUID`.

## Checklist

 - [x] Link to issue this PR refers to (if applicable): Fixes #829


CHANGES.rst
```diff
@@ -5,6 +5,14 @@ Changes for crate
 Unreleased
 ================
 
+- Breaking change: ``DefaultTypeConverter`` now decodes ``DataType.UUID``
```
florinutz: I think this is more of an addition (e.g. "``DefaultTypeConverter`` decodes ``UUID`` columns to ``uuid.UUID``), than a breaking change

florinutz (approved):
good stuff!

src/crate/client/converter.py
```diff
@@ -122,19 +136,33 @@ class DataType(Enum):
...
+    FLOAT_VECTOR = 28
+    UUID = 29
+    REGTYPE = 30
     ARRAY = 100
 
 
 ConverterMapping = Dict[DataType, ConverterFunction]
 
 
+def _resolve(type_: Any) -> Optional[DataType]:
```
mfussenegger:
```suggestion
def _resolve(type_: int) -> Optional[DataType]:
```

mfussenegger (approved):

src/crate/client/converter.py
bgunebakan: Thanks, I also added type checking on Converter class because mypy was complaining.

CHANGES.rst
bgunebakan: I labelled it breaking because a column migrated from text to uuid starts handing back `uuid.UUID` instead of `str`. users could be need to do code changes for new python type.

PR #832, opened 2026-09-16
Release crate-python v2.3.0

### LINKED ISSUE #792: Validate packaging configuration: `uv build` vs. `hatch build`
<blockquote>

Sorry to nitpick further here, but I'm a bit confused about this change:

> Using `uv build` creates a wheel package without any code. (8200dfeb14818e9dbbc37737aef67cb7f33ad22b)

Isn't this an indication that there's something wrong with the pyproject.toml definition? `uv build` acts as a build frontend, using the build backend defined in `[build-system]`. So `uv build` should actually install & use hatch itself.

</blockquote>

_Originally posted by @mfussenegger in https://github.com/crate/crate-python/issues/786#issuecomment-3996743867_

### LINKED ISSUE #831: Release 2.3.0
### Added

- CrateDB column type identifiers in `DataType`: `INTERVAL` (17), `ROW` (18), `FLOAT_VECTOR` (28), `UUID` (29), and `REGTYPE` (30). (#830)

### Fixed

- `Converter.get()` raised `ValueError` for column type identifiers it does not know. Unknown identifiers now fall back to the default converter, so queries on columns using newer CrateDB types no longer fail. (#830)

### Behavior changes

- `connect()` now raises `ConnectionError` immediately if no configured server node responds, instead of deferring the error until the first statement is executed. Code that creates a connection eagerly, before the cluster is reachable, needs to handle the error at `connect()` time. (#711)
- `DefaultTypeConverter` now decodes `DataType.BIT` columns, converting CrateDB's `B'0110'` wire format to a plain string of `0`/`1` digits. Code that stripped the wrapper itself needs to be adjusted, or can restore the previous behaviour by mapping `DataType.BIT` to a converter of its own. (#828)
- `DefaultTypeConverter` now decodes `DataType.UUID` columns to Python `uuid.UUID` objects instead of returning the raw string. Code that expected a string can call `str()` on the value, or map `DataType.UUID` to a converter of its own.

Both converter changes only affect code that opts into `DefaultTypeConverter`; the default `Converter` is unchanged.

bgunebakan:
- Added ``DefaultTypeConverter`` support that decodes ``DataType.UUID`` columns to Python ``uuid.UUID`` objects instead of returning the raw string.
- Added CrateDB column type identifiers to ``DataType``: ``INTERVAL`` (17), ``ROW`` (18), ``FLOAT_VECTOR`` (28), ``UUID`` (29), and ``REGTYPE`` (30). Fixed ``Converter.get()`` raising ``ValueError`` for column type identifiers it does not know. Unknown identifiers now fall back to the default converter.

- Breaking change: ``connect()`` now raises ``ConnectionError`` immediately if no configured server node responds.

- Breaking change: ``DefaultTypeConverter`` now decodes ``DataType.BIT`` columns, converting CrateDB's ``B'0110'`` wire format to a plain string of ``0``/``1`` digits. Code that stripped the wrapper itself needs to be adjusted, or can restore the previous behaviour by mapping ``DataType.BIT`` to a converter of its own.

## Fix for packaging
Our sdist include pattern was /src/crate/*.py and published sdists contained no source at all. This already reported on @mfussenegger's comment in #786 and mentioned in #792. I explored while releasing a beta package. Fixed by including `/src/crate` as a directory. Also anchored the `*.rst/*.txt` patterns, which matched at any depth and swept unrelated files into the sdist. I keep this fix here because it's related with package. 

## Checklist

 - [x] Link to issue this PR refers to (if applicable): Fixes #831 
 - [x] Also Closes #792


florinutz (approved):
looks good, let's release 🥇

PR #834, opened 2026-09-18
Add `probe` option to connect without a responding server

### LINKED ISSUE #833: Allow `connect()` (again) while the server is not responding
Since 2.3.0, `connect()` contacts the configured servers and raises `ConnectionError` when none of them answers (#711). For most programs that is an improvement: the failure arrives where the connection is made, not at the first statement.

A test suite that talks to a cluster while nodes are starting or shutting down now fails inside `connect()`. crate's blackbox tests hit that and crate pinned `crate<2.3` to unblock CI (see crate/crate#20233). The same applies to programs that connect while the server is still coming up (e.g. a CI service container).

Proposal: `connect(..., probe=False)` skips the check at construction. The server version is then resolved when something first reads `lowest_server_version`, and an unreachable cluster fails at the first statement, **as it did before 2.3.0**. The default stays `probe=True`, so nothing changes for callers who want the early failure.

florinutz:
With this changeset, `connect()` now takes a `probe` argument that defaults to `True`. In this case, the servers are contacted while the connection is created, and `ConnectionError` is raised when none of them responds. This would be **what 2.3.0 does**.

With `probe=False`, creating the connection performs no request. `lowest_server_version` is resolved when something first reads it, and reading it raises `ConnectionError` while no server responds. A caller that never reads it talks to the cluster on the first statement, **as before 2.3.0**.

This gives callers (that talk to a cluster which is starting or shutting down) a way back. crate's blackbox tests pinned `crate<2.3` for that reason in crate/crate#20233, and CI service containers have the same race. So these callers would pass `probe=False` to every connect() that causes troubles, and this would just use the previous behavior instead of the one introduced in 2.3.0.

Closes #833.


matriv:
@florinutz Thx for the quick workaround!

Meanwhile we've fixed things on the crate side with https://github.com/crate/crate/pull/20244 and https://github.com/crate/crate-qa/pull/451, so for us there is no need to proceed with this change.
But, maybe it's meaningful for other scenarios, in which case feel free to proceed with this PR.

